PR #18, opened 2017-09-03
Deduplicate nodes when creating new configurations

s111:
Manager.NewConfiguration returns a configuration with `node1,node2,node2` given ids `id1,id2,id2`.
I would like to change this function call to return a configuration with `node1,node2` so that it's easier to make idempotent add operations.
(Alternatively, disallow duplicate node ids and return an error as is done with gorums.NewManager.)

### DIFF
--- a/dev/mgr.go
+++ b/dev/mgr.go
@@ -227,7 +227,16 @@ func (m *Manager) NewConfiguration(ids []uint32, qspec QuorumSpec) (*Configurati
 	}
 
 	var cnodes []*Node
+	unique := make(map[uint32]struct{})
+	var deduped []uint32
 	for _, nid := range ids {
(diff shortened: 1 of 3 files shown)


dev/mgr.go
```diff
@@ -227,7 +227,16 @@ func (m *Manager) NewConfiguration(ids []uint32, qspec QuorumSpec) (*Configurati
 	}
 
 	var cnodes []*Node
+	unique := make(map[uint32]struct{})
```
tormoder: Minor, but if you want you can use

```go
duplicate := make(map[uint32]bool)
```

as the set, and

```go
if unique[nid] {
```

but then perharps use another name for `unique`.

Issue #33, opened 2017-09-13
Should we rename <MethodName>Reply type to something else?


meling:
The following is just a small excerpt from a longer discussion in [relab/gorums-dev/#29](https://github.com/relab/gorums-dev/issues/29). The below description captures the main outstanding issue that haven't been decided yet. Namely what to name the reply type, when it is different from type specified in the proto file, because we need to return a more sophisticated data structure, for call types such as correctables and futures.

**Below text is copied from the old issue:**
Note that, I'm not happy with the suffix `Reply` to the `ReadCorrectableReply` which comes from the `ReadCorrectable` method name.
```
sm.TypeName = sm.MethodName + "Reply"
```
Ignoring the fact that in the current example, I've named the method `ReadCorrectable`, perhaps it would be better to name the different reply types with their respective names, e.g. `Correctable`. That is, for a `ReadCorrectable` method, the reply type would become `ReadCorrectableCorrectable`.

This looks ugly for our example case, but if our method name is just `Read`, then the return type becomes `ReadCorrectable` and `ReadFuture` and so on.

Another alternative (that was used in a previous implementation) is to just use the method name as the name for the reply type.

Anyone have any input on what is the preferred choice on this naming??

meling:
Would it make sense to use
```go
sm.TypeName = sm.CustomRespName + "Wrapper"
```
Note that `sm.CustomRespName` is set to `sm.RespName` if `sm.CustomRespName` is the empty string.

This would give us reply types such as `StateWrapper` and `ReadRequestWrapper`.

However, I'm not sure if this unique, e.g. if multiple methods return `State`.

meling:
Closed by 35059f7

PR #34, opened 2017-09-19
Rearchitected the datatypes to be more appropriately named.

meling:
At the same time, this approach allows for reuse of internal datatypes.

I have come up with a solution that I like. Basically, it separates out the datatypes in a new template file and allows to reuse the internal datatypes across different methods; these are named `internalState` and `internalWriteResponse` in our current examples. Furthermore, it names the exported datatypes by call type, e.g. you can expect to see `CorrectableState` and `FutureWriteResponse` and such. However, there will not be a `QuorumCallState` since quorum calls only returns a `State` object.

This should solve issue #33. There may be room for some minor tweaks and general improvements that I may get to tomorrow or later in the week.


plugins/gorums/gorums.go
```diff
@@ -335,6 +335,20 @@ type responseType struct {
 	Multicast          bool
 }
 
+func NewResponseType(respType string, sm *serviceMethod) responseType {
```
tormoder: Unexport this function?

plugins/gorums/ext.go
```diff
@@ -7,17 +7,17 @@ import (
 	"github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
 )
 
-func qcName() string {
-	return gorumsproto.E_Qc.Name
-}
+// func qcName() string {
```
tormoder: Delete if they are no longer used?

Issue #35, opened 2017-09-24
Gorums cannot handle nil for non-PerNodeArg Gorums RPC calls.


wruiwr:
For my read-write distributed storage implemented by Gorums,
if I send a nil message through either read or write quorum call,
then the quorum call returns a nil reply with error nil.

The reason for this issue is the callGRPC methods generated by Gorums.

For example in method: callGRPCWrite:

```go
if arg == nil {
    // send a nil reply to the for-select-loop
    replyChan <- internalWriteResponse{node.id, nil, nil}
    return
}
```

This allows the write quorum call to send a nil write request and reply a nil as the write response with error message nil. (the quorum function can receive enough replies to obtain a quorum)

So, in my quorum functions, I currently wrote some code to handle the nil replies, otherwise, it can cause panic, since there's no value and timestamp to get if replies are nils.

s111:
I'm not sure I understand the issue here.
gRPC doesn't allow passing nil, as the marshalling step will panic (as it should).
It seems that we now protect against passing nil values on all methods generated by Gorums ( #12).

I think that passing nil values should panic, as this is the developers mistake.
There is however one special case where I think we should handle nil values: when a per-node-arg function returns nil, we should drop that request and reduce the expected number of replies for that call. This would allow us to do some cool things with the per-node-arg functions.
What do you think @meling?

wruiwr:
The quorum call actually can invoke gRPC calls with nil message and the quorum functions can also receive enough replies (also nil) to obtain a quorum. 

So, the returned reply for this quorum call in this case is nil, however if the server failure happens, then no quorum is obtain, the returned reply is also nil. Therefore, I think in my tests, this two cases return the same values (both nils)  (one is not an error (the err is nil) according to Hein, another is caused by the server failure), so I can not check the difference.

wruiwr:
I think Hein has a solution about this issue. ;)

s111:
A quorum call works with nil values as we circumvent the actual gRPC call:
```go
if arg == nil {
	       // send a nil reply to the for-select-loop
	      replyChan <- internalWriteResponse{node.id, nil, nil}
	      return
}
```
An actual gRPC call would panic with:
```
panic: rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil
```

[...]

PR #36, opened 2017-09-24
Fix nil values being added to the reply slice

s111:
TestPerNodeArg is left failing to demonstrate the issue.
The test should fail while in `master` is succeeds.
I'll fix it when we agree whether this is a good solution.
See #35.

### DIFF
--- a/dev/calltype_correctable_stream_gen.go
+++ b/dev/calltype_correctable_stream_gen.go
@@ -120,7 +120,8 @@ func (c *Configuration) readCorrectableStream(ctx context.Context, a *ReadReques
 		}()
 	}
 
-	replyChan := make(chan internalState, c.n)
+	expected := c.n
+	replyChan := make(chan internalState, expected)
 	for _, n := range c.nodes {
 		go callGRPCReadCorrectableStream(ctx, n, a, replyChan)
 	}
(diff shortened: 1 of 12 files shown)


dev/calltype_correctable_stream_gen.go
```diff
@@ -120,7 +120,8 @@ func (c *Configuration) readCorrectableStream(ctx context.Context, a *ReadReques
 		}()
 	}
 
-	replyChan := make(chan internalState, c.n)
+	expected := c.n
```
tormoder: I haven't looked at this PR closely yet.

But this code is not needed? (It does nothing new/change the logic).
Maybe you didn't mean this intentionally since it's a result of the code generation.

And for stream correctables there are no notion of "expected", since the servers can return an arbitrary number of replies.

meling: The `expected` variable is technically only needed for the `per_node_arg` option, but it is perhaps easier to just keep it for both cases to avoid yet another `{{if .PerNodeArg}}` conditional in the template code. To understand when it is needed you need to look at the generated `writePerNode` function in `calltype_quorumcall_gen.go`.

tormoder: Yes, my comment was only targeted at the correctables and not in general.

meling:
Merging this into master now. Further changes can be done on master, if there are remaining issues.

PR #43, opened 2018-02-26
simplified String() method to ease debugging

meling:
added FullString() method for those cases where more details are
needed.

### DIFF
--- a/dev/node_func.go
+++ b/dev/node_func.go
@@ -33,12 +33,22 @@ func (n *Node) Port() string {
 }
 
 func (n *Node) String() string {
-	n.mu.Lock()
-	defer n.mu.Unlock()
-	return fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


dev/node_func.go
```diff
@@ -33,12 +33,22 @@ func (n *Node) Port() string {
...
-	n.mu.Lock()
-	defer n.mu.Unlock()
-	return fmt.Sprintf(
-		"node %d | addr: %s | latency: %v",
-		n.id, n.addr, n.latency,
-	)
+	if n != nil {
+		return fmt.Sprintf("addr: %s", n.addr)
+	}
+	return ""
```
tormoder: Maybe `return "<nil>"` to be explicit about that the `n` is nil?
An empty string is easy to miss in the output.
Ditto for line 51.

PR #49, opened 2018-07-11
replaced deprecated grpc.WithTimeout() for use in the tests

meling:
this was a fairly extensive change since we pass in these grpc
options into the NewManager() function, and we use this a lot.


meling:
@tormoder I merged this already.

PR #51, opened 2020-01-13
generate a ".gorums.go" file instead of including gorums code in protobuf ".pb.go" file

Raytar:
With this change, the gorums plugin will generate two files: 
*  A `.gorums.go` file containing all of the gorums code
* A `.pb.go` file containing protobuf code.

This fixes the problem where protobuf renames the imports that gorums needs.


meling:
I did some tests building with a new proto file with message types in a separate proto file and importing that proto file; turns out this wasn't trivial to solve because the imported proto file does not get imported in the `.gorums.go` file. (Not sure if this used to work before, but I think Sebastian's Raft implementation does use imports quite a bit... need to check.)

Issue #16, opened 2017-09-01
Quorum calls invoked in-order might send messages to individual servers out-of-order


meling:
![qrpc](https://cloud.githubusercontent.com/assets/68231/25727096/a3416ba0-3127-11e7-8a2b-76b987ce87c2.png)

See the above sketch illustrating an important problem with synchronous quorum calls.

**The details:**
When a client performs what she thinks to be a synchronous quorum call by sending an RPC to all servers in a configuration, it is actually spawning a goroutine for each server. This is illustrated below for a `Read` quorum call.

```go
func (c *Configuration) Read(ctx context.Context, a *ReadRequest) (resp *State, err error) {
	...
	for _, n := range c.nodes {
		go callGRPCRead(ctx, n, a, replyChan)
	}
	...
```

```go
func callGRPCRead(ctx context.Context, node *Node, arg *ReadRequest, replyChan chan<- readReply) {
	err := grpc.Invoke(ctx, "/dev.Register/Read", arg, reply, node.conn)
	replyChan <- readReply{node.id, reply, err}
}
```

This `Read` quorum call can complete before all replies have been received, because only a subset of successful replies are needed to complete a quorum call. Exactly when enough replies have been received is determined by a quorum function, I won't explain the details here, except to say that in a typical majority quorum function, invoking a configuration with three servers, a majority of two replies is enough to complete a call. This has the potential to leave some goroutines running, or possibly even waiting to be started.

Thus, in rare cases it can happen that synchronous quorum call QC1 completes, but some of its goroutines haven't even started running yet. Moreover, since QC1 is complete, the client can perform the next quorum call QC2, and some of its goroutines may actually be scheduled before those belonging to QC1.

Note that it is okay for the goroutines of QC1 to linger after QC2 has started, as long as the RPC calls (TCP send call) belonging to QC1 are invoked before those belonging to QC2 for the same server.

Furthermore, often it is undesirable to cancel these lingering goroutines since it may cause some (slow) servers to become outdated with the rest. But there are some applications/protocols, where canceling these goroutines would be okay. However, that should not be the default since those that wish to cancel can do so through the context `ctx`.

**The problem it causes:**
A server can receive a message out-of-order, even though the underlying networking takes place over TCP. This is unexpected behavior.

**Is this relevant in practice?**
[...]

PR #52, opened 2020-02-12
Add a new quorum call option with strict ordering

Raytar:
This pull request adds a new quorum call option, `qc_strict_ordering`, that has strict ordering of messages due to its use of streams. The code is being used for an implementation of HotStuff in https://github.com/relab/hotstuff (Currently a private repo, not sure who has access).

The code is similar to the stream based approach suggested by @meling in [#16](https://github.com/relab/gorums/issues/16#issuecomment-390379368). The major difference is that each node runs two goroutines on `Node.connect()`:

* A goroutine that takes messages from the Node's `Send` channel and sends them on the stream.
* A goroutine that receives messages from the stream and puts the messages on a channel corresponding to the ID of the message.

The reason for using these two goroutines per Node is that it avoids waiting for a response. In other words, multiple messages can be sent to a Node independently of how many replies it has sent back.

Because of this behavior it is necessary to give each request/response pair a unique ID. I did not find a clever way to embed this ID field in the messages. Thus it is currently required that the user specifies the name of a common field in both the request and response fields of a method with an unsigned integer type that can be used by gorums. See the example below:

```protobuf
service OrderedStorage {
	// WriteOrdered is a synchronous quorum call that enforces a strict
	// ordering of the messages that are sent to a node, such that
	// a quorum call Q1 will deliver its messages to the nodes
	// respectively before a later quorum call Q2 delivers its messages.
	// The order of replies is not guaranteed for concurrent quorum calls.
	rpc WriteOrdered(stream State) returns (stream WriteResponse) {
		// The option specifies a common field in the `State` and `WriteResponse`
		// messages of type an unsigned integer type that gorums can use to keep
		// track of which requests and replies belong together.
		option (gorums.qc_strict_ordering) = "GorumsMessageID";
	}
}

message State {
	string Value 	= 1;
	int64 Timestamp = 2;
	uint32 GorumsMessageID = 3;
}

message WriteResponse {
	bool New = 1;
	uint32 GorumsMessageID = 2;
}
```

A counter in `Manager` is used to generate the message IDs.
Additionally, a map of message IDs to channels is used to provide a way for Nodes to send the response messages back to the quorum call function.

A helper function is generated to simplify the server implementation:
```go
// StorageServerBasic implements the GRPC server API
// WriteOrdered implements the RPC `WriteOrdered`
func (s *StorageServerBasic) WriteOrdered(srv Storage_WriteOrderedServer) error {
	return WriteOrderedServerLoop(srv, func(req *State) *WriteResponse {
		// process state and return a *WriteResponse
	})
}
```


Raytar:
Closing this PR in favor of #55

PR #53, opened 2020-03-16
Don't embed streams in "nodeServices"

Raytar:
Embedding will not work when there are multiple quorum calls that use streams.


meling:
Yep. That's my bad; didn't test it with multiple multicast/stream methods. Could you add another multicast method, e.g. `WriteMulticast()`, to the `zorums.proto` file as a "test case"?

Raytar:
Done.

Raytar:
Fixed a typo.

meling:
LGMT

PR #54, opened 2020-03-17
Use a context that doesn't timeout for creating streams

Raytar:
The streams should stay up until the Node is closed.


meling:
Ups. Good catch.

PR #55, opened 2020-03-27
Add a strict ordering option

Raytar:
This pull request adds a new option: `gorums.strict_ordering`, that uses a single gRPC stream per node to send all "strict ordering" messages. The strict ordering is achieved due to the fact that gRPC streams preserve the order in which messages are sent on the stream.

The code is mostly based on my previous pull request (#52), but modified such that there is only one gRPC stream per Node instead of one gRPC stream per "strict ordering" method. The reason for this is to avoid the requirement that each "strict ordering" message must specify an "ID" field. Another benefit is that calls across different strict ordering methods is preserved. Let's consider the example below:

```protobuf
service GorumsTest {
	rpc QC(Request) returns (Response) {
		option (gorums.qc) = true;
		option (gorums.strict_ordering) = true;
	}

	rpc UnaryRPC(Request) returns (Response) {
		option (gorums.strict_ordering) = true;
	}
}

message Request {
	uint64 Num = 1;
}

message Response {
	bool InOrder = 1;
}
```

In order to use strict ordering, the `gorums.strict_ordering` is set in combination with other gorums options to determine the call type of the method. The `QC` rpc is a `quorum call` since it uses the `gorums.qc` option. The `UnaryRPC` rpc is really just a regular rpc, but it is also sent over the same gRPC stream as `QC`, since it also specifies the `gorums.strict_ordering` option. Note that a "message ID" field is not required unlike the previous version (PR #52), and the rpc do not specify its messages as streams.

Again, its important to note that the order of calls to both `QC` and `UnaryRPC` in sequence is preserved.

### API

The client API for Quorum Calls should be identical to the regular Quorum Calls. The client API for Unary RPCs is similar to the gRPC way, with the method being exposed through the Nodes. The server API is quite different, as the below example shows:

```go
type testSrv struct {}

// QC is the server-side handler for the 'QC' Quorum Call.
func (s *testSrv) QC(req *Request) *Response {
  ...
}

// UnaryRPC is the server-side handler for the 'UnaryRPC' RPC.
func (s *testSrv) UnaryRPC(req *Request) *Response {
  ...
}

func main() {
  tSrv := &testSrv{}

  srv := NewGorumsServer()

  // each handler is registered with the server
  srv.RegisterQCHandler(tSrv)
  srv.RegisterUnaryRPCHandler(tSrv)
  
  lis, _ := net.Listen(...)

  go srv.Serve(lis)

  ...

  srv.Stop()
}
```

### Implementation details

Under the hood, the strict ordering methods marshal messages into a `google.protobuf.Any` format and sends it to the Nodes through each Node's `sendQ` channel. The Node takes these messages in sequence and sends them on the stream. Messages are sent back through a `recvQ` channel stored in a map for each message ID. The message ID's are part of the gorums message types sent on the stream. The definition of the gorums messages and the "strict ordering stream" is shown below:

```protobuf
service Gorums {
  rpc StrictOrdering(stream Message) returns (stream Message);
}

message Message {
  uint64 ID = 1;
  string URL = 2;
  google.protobuf.Any Data = 3;
}
```

Some state is stored both in the Manager and Nodes. This state includes the recvQ channels map, the sendQ channels, a counter for the message ID, and some mutexes. This is all quite similar to pr #52, except that the state is shared for all strict ordering methods.

In order to correctly generate these new "calltypes" that rely on a combination of `gorums.strict_ordering` + other options, I created some internal only options that the plugin uses to detect the correct strict ordering type. These are located in `internal/strictordering` and are only meant to be used internally by the `protoc-gen-gorums` program when it decides which templates to use. This is a bit of a hack, but I think that the simplicity of being able to specify `gorums.strict_ordering` + other options is nicer than creating new options for each "strict ordering" variant.

### Other notes

* Because the `StrictOrdering` rpc is the only "real" gRPC call that is made when using strict ordering methods, it is not necessary to compile a gRPC file for a service that only uses strict ordering methods.

* Reconnection has been added to strict ordering methods with this PR, and is configurable through the `WithBackoff` manager option that takes a gRPC `BackoffConfig`. This option also sets the backoffconfig for the regular gRPC connections. While reconnection is in progress, all strict ordering rpcs will fail with an error.


Raytar:
I added some tests to `cmd/protoc-gen-gorums/tests/strictordering` is that the correct place to put them? Also, for some reason VSCode is showing a lot of "undeclared name" errors in the test file, but the tests run just fine. Probably a bug with vscode-go.

meling:
Yes, this a known bug in [gopls](https://github.com/golang/go/issues/32394), which VSCode use.

About the placement of the tests. My current plan is to keep things that only refer to testing of the plugin and generated code inside `cmd/protoc-gen-gorums/tests`, while other tests that may involve some form of integration and running of servers and clients, I'm thinking that could be moved to a top level tests folder or examples folder. But I've not written any myself, so I'm still thinking about the best way to make this follow standard Go practices and so forth.

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+	nid   uint32
+	reply *anypb.Any
+	err   error
+}
+
+type requestHandler func(*gorums.Message) *gorums.Message
+
+type strictOrderingManager struct {
+	msgID    uint64
+	recvQ    map[uint64]chan *strictOrderingResult
```
meling: Consider if perhaps [`sync.Map`](https://golang.org/pkg/sync/#Map) should be used instead, because:
_The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex._

It would be interesting to do a performance comparison, given the comment from the documentation, which seems to match our use case, since we only ever write to the same key once, but read it many times.

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+
+type strictOrderingResult struct {
+	nid   uint32
+	reply *anypb.Any
+	err   error
+}
+
+type requestHandler func(*gorums.Message) *gorums.Message
+
+type strictOrderingManager struct {
```
meling: Don't think this struct's name is representative for its content. What about `receiveBuffer` or `receiveQueue`?

Also, I propose to augment its API with functions for safely accessing the map, e.g. Load/Store/Delete. With this API, you could either forward to the corresponding `sync.Map` API or encapsulate snippets like this:
```go
	s.recvQMut.RLock()
	if c, ok := s.recvQ[resp.GetID()]; ok {
		c <- &strictOrderingResult{nid: s.node.ID(), reply: resp.GetData(), err: nil}
	}
	s.recvQMut.RUnlock()
```

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+func newStrictOrderingManager() *strictOrderingManager {
+	return &strictOrderingManager{
+		recvQ: make(map[uint64]chan *strictOrderingResult),
+	}
+}
+
+func (m *strictOrderingManager) createStream(node *Node, backoff backoff.Config) *strictOrderingStream {
+	return &strictOrderingStream{
+		node:      node,
+		nextMsgID: m.nextMsgID,
```
meling: Shouldn't this read of `m.nextMsgID` be atomic?

meling: Ah, this is a func().

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+	recvQMut sync.RWMutex
+}
+
+func newStrictOrderingManager() *strictOrderingManager {
+	return &strictOrderingManager{
+		recvQ: make(map[uint64]chan *strictOrderingResult),
+	}
+}
+
+func (m *strictOrderingManager) createStream(node *Node, backoff backoff.Config) *strictOrderingStream {
```
meling: Would it not be better to have this as a method on `Node` and pass `m *receiveQueue` as an argument.

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+	gorumsClient gorums.GorumsClient
+	gorumsStream gorums.Gorums_StrictOrderingClient
+	streamMut    sync.RWMutex
+	streamBroken bool
+	sendQ        chan *gorums.Message
+	recvQ        map[uint64]chan *strictOrderingResult
+	recvQMut     *sync.RWMutex
+}
+
+func (s *strictOrderingStream) connect(ctx context.Context, conn *grpc.ClientConn) error {
```
meling: If this were named `connectGorumsStream` (or similar), you could use struct embedding in `Node` like this:
```go
	*strictOrderingStream
```
Instead of:
```go
	strictOrdering *strictOrderingStream
```

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
+package dev
+
+import (
+	"context"
+	"math"
+	"math/rand"
+	"sync"
+	atomic "sync/atomic"
```
meling: Nitpick: Avoid named imports for atomic and status below.

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -84,6 +95,26 @@ func genGorumsMethod(g *protogen.GeneratedFile, method *protogen.Method) string
...
+		for _, method := range service.Methods {
+			if hasStrictOrderingOption(method, methodOptions...) {
+				fmt.Fprintf(os.Stderr, "processing %s\n", method.GoName)
+				g.P(genStrictOrderingMethod(g, method))
+			}
+		}
+	}
+}
+
+func genStrictOrderingMethod(g *protogen.GeneratedFile, method *protogen.Method) string {
```
meling: This func is almost identical to `genGorumsMethod`. Please use this func instead. The method name in the error message should uniquely identify where the problem is. Reaching the panic can only happen if someone is playing around with adding more options and forgot to update some of the structs.

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -62,12 +65,20 @@ func GenerateFileContent(gen *protogen.Plugin, file *protogen.File, g *protogen.
 	g.P()
 	genGorumsMethods(data, gorumsCallTypes...)
 	g.P()
+	// generate all strict ordering methods
+	genStrictOrderingMethods(servicesData{g, file.Services}, strictOrderingCallTypes...)
```
meling: `s/servicesData{g, file.Services}/data/`

cmd/protoc-gen-gorums/gengorums/gorums_func_map.go
```diff
@@ -225,3 +232,24 @@ func mustExecute(t *template.Template, data interface{}) string {
 	}
 	return b.String()
 }
+
+func numStrictOrderingMethods(services []*protogen.Service) int {
```
meling: Nitpick: maybe better to use `hasStrictOrderingMethods`, which could return a bool.

cmd/protoc-gen-gorums/gengorums/template_globals.go
```diff
@@ -0,0 +1,5 @@
+package gengorums
+
+var globals = `
+const numStrictOrderingMethods = {{numStrictOrderingMethods .Services}}
```
meling: Nitpick: maybe better to use `hasStrictOrderingMethods`.

gorums.proto
```diff
@@ -14,6 +15,17 @@ extend google.protobuf.MethodOptions {
...
   string custom_return_type = 51000;
 }
+
+service Gorums {
+  rpc StrictOrdering(stream Message) returns (stream Message);
+}
+
+message Message {
+  uint64 ID = 1;
+  string URL = 2;
```
meling: This should be called `Method`, following gRPC's naming.

gorums.proto
```diff
@@ -14,6 +15,17 @@ extend google.protobuf.MethodOptions {
   bool qc_future = 50004;
   bool qf_with_req = 50005;
   bool per_node_arg = 50006;
+  bool strict_ordering = 50007;
 
   string custom_return_type = 51000;
 }
+
+service Gorums {
```
meling: On second thought, I'm not sure we want to expose this RPC and `Message` here, since they are internal to Gorums; should not be used by others. I think I might have suggested this, but you know... Seems like these would fit well in the `strictordering.proto`, or am I missing something?

cmd/protoc-gen-gorums/gengorums/gorums_dev.go
```diff
@@ -10,7 +10,8 @@ import (
 
 // GenerateDevFile generates a _{{gorumsType}}_gorums.pb.go file containing Gorums service definitions.
 func GenerateDevFile(gorumsType string, gen *protogen.Plugin, file *protogen.File) {
-	if len(file.Services) == 0 || !hasGorumsType(file.Services, gorumsType) {
+	if len(file.Services) == 0 || !hasGorumsType(file.Services, gorumsType) &&
+		!hasStrictOrderingType(file.Services, gorumsType) {
```
meling: Remove this && expression and add the following at the end of `hasGorumsType`:
```go
	return hasStrictOrderingType(services, gorumsType)
```

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -117,8 +148,24 @@ func hasGorumsType(services []*protogen.Service, gorumsType string) bool {
 	return false
```
meling: Replace this `return false` with:
```go
	return hasStrictOrderingType(services, gorumsType)
```
(see also companion comment in gorums_dev.go:14)

cmd/protoc-gen-gorums/dev/server.go
```diff
@@ -0,0 +1,71 @@
...
+	handlers map[string]requestHandler
+}
+
+func newStrictOrderingServer() *strictOrderingServer {
+	return &strictOrderingServer{
+		handlers: make(map[string]requestHandler),
+	}
+}
+
+func (s *strictOrderingServer) registerHandler(url string, handler requestHandler) {
```
meling: I think `url` should be named `method` to follow gRPC's naming scheme.

cmd/protoc-gen-gorums/dev/node.go
```diff
@@ -37,15 +41,26 @@ func (n *Node) connect(opts managerOptions) error {
 	if err != nil {
 		return fmt.Errorf("dialing node failed: %w", err)
 	}
-	return n.connectStream() // call generated method
+	// a context for all of the streams
+	ctx, n.cancel = context.WithCancel(context.Background())
+	// only start strictOrdering RPCs when needed
+	if numStrictOrderingMethods > 0 {
```
meling: Nitpick: Make bool?

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+type strictOrderingResult struct {
+	nid   uint32
+	reply *anypb.Any
+	err   error
+}
+
+type requestHandler func(*gorums.Message) *gorums.Message
```
meling: Should this be moved to `dev/server.go`? Would also be useful to have a comment associated.

cmd/protoc-gen-gorums/dev/strictordering.go
meling: Something like this:
```go
func (node *Node) createStream(m *strictOrderingManager, backoff backoff.Config) {
	node.strictOrdering = &strictOrderingStream{
		node:      node,
		nextMsgID: m.nextMsgID,
		backoff:   backoff,
		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
		sendQ:     make(chan *gorums.Message),
		recvQ:     m.recvQ,
		recvQMut:  &m.recvQMut,
	}
}
```

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+func (s *strictOrderingStream) reconnectStream(ctx context.Context) {
+	s.streamMut.Lock()
+	defer s.streamMut.Unlock()
+
+	var retries float64
+	for {
+		var err error
+		s.gorumsStream, err = s.gorumsClient.StrictOrdering(ctx)
+		if err == nil {
+			s.streamBroken = false
```
meling: Should this err be recorded?:
```go
	s.node.setLastErr(err)
```

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+		// else try to send message
+		s.streamMut.RLock()
+		err := s.gorumsStream.SendMsg(req)
+		if err == nil {
+			s.streamMut.RUnlock()
+			continue
+		}
+		s.streamBroken = true
+		s.streamMut.RUnlock()
+		// return the error
```
meling: Should this record the err?:
```go
	s.node.setLastErr(err)
```

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+	go s.recvMsgs(ctx)
+	return nil
+}
+
+func (s *strictOrderingStream) sendMsgs() {
+	for req := range s.sendQ {
+		// return error if stream is broken
+		if s.streamBroken {
+			err := status.Errorf(codes.Unavailable, "stream is down")
+			s.recvQMut.RLock()
```
meling: Should this err be recorded?
```go
	s.node.setLastErr(err)
```

cmd/protoc-gen-gorums/dev/opts.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	backoff "google.golang.org/grpc/backoff"
```
meling: no need for named import

gorums.proto
Raytar: It's "internal" to gorums, but imported by the generated code, so I don't think it's possible to put it in an "internal" package.

gorums.proto
```diff
@@ -14,6 +15,17 @@ extend google.protobuf.MethodOptions {
...
+  rpc StrictOrdering(stream Message) returns (stream Message);
```
meling, line 24: Name suggestion: `StrictOrdering` --> `NodeStream`

gorums.proto
```diff
@@ -14,6 +15,17 @@ extend google.protobuf.MethodOptions {
...
+}
+
+message Message {
```
meling, line 27: So, since this is probably moved into `dev` (i.e. made part of the Gorums generated code), maybe it would be good to name this something else too, e.g. `Message` --> `GorumsMessage` or `InternalMessage` or something better...

cmd/protoc-gen-gorums/dev/strictordering.go
```diff
@@ -0,0 +1,165 @@
...
+		recvQ:     m.recvQ,
+		recvQMut:  &m.recvQMut,
+	}
+}
+
+func (m *strictOrderingManager) nextMsgID() uint64 {
+	return atomic.AddUint64(&m.msgID, 1)
+}
+
+type strictOrderingStream struct {
```
meling, line 55: Maybe this could be named `orderedNodeStream`

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -62,12 +65,20 @@ func GenerateFileContent(gen *protogen.Plugin, file *protogen.File, g *protogen.
...
 
 func genGorumsMethods(data servicesData, methodOptions ...*protoimpl.ExtensionInfo) {
 	g := data.GenFile
 	for _, service := range data.Services {
 		for _, method := range service.Methods {
+			// skip all strict ordering methods.
+			// they will be generated by genStrictOrderingMethods
+			if hasMethodOption(method, gorums.E_StrictOrdering) {
+				continue
+			}
```
meling, line 81: Together with other changes (to be suggested), the loop body can be replaced with this:
```go
	if hasMethodOption(method, gorums.E_StrictOrdering) {
		if hasStrictOrderingOption(method, methodOptions...) {
			fmt.Fprintf(os.Stderr, "processing %s\n", method.GoName)
			g.P(genGorumsMethod(g, method))
		}
	} else if hasMethodOption(method, methodOptions...) {
		fmt.Fprintf(os.Stderr, "processing %s\n", method.GoName)
		g.P(genGorumsMethod(g, method))
	}
```
(this will allow to eliminate `genStrictOrderingMethods`, which is very similar to `genGorumsMethods`...)

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -84,6 +95,26 @@ func genGorumsMethod(g *protogen.GeneratedFile, method *protogen.Method) string
 	panic(fmt.Sprintf("unknown method type %s\n", method.GoName))
 }
 
+func genStrictOrderingMethods(data servicesData, methodOptions ...*protoimpl.ExtensionInfo) {
```
meling, line 98: Can be eliminated / replaced by `genGorumsMethod`.

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -203,6 +277,27 @@ func hasMethodOption(method *protogen.Method, methodOptions ...*protoimpl.Extens
...
+			return false
+		}
+	}
+	return true
+}
+
+// hasStrictOrderingOption returns true if the method has one of the given strict ordering method options.
+func hasStrictOrderingOption(method *protogen.Method, methodOptions ...*protoimpl.ExtensionInfo) bool {
+	for _, option := range methodOptions {
+		if strictOrderingTypeCheckers[option](method) {
```
meling, line 294: This will fail if one of the provided `methodOptions` is not in the map. Replace with:
```go
	if f, ok := strictOrderingTypeCheckers[option]; ok && f(method) {
```

cmd/protoc-gen-gorums/gengorums/gorums_dev.go
```diff
@@ -40,6 +43,8 @@ func GenerateDevFile(gorumsType string, gen *protogen.Plugin, file *protogen.Fil
 	default:
 		if methodOption, ok := gorumsTypes[gorumsType]; ok {
 			genGorumsMethods(data, methodOption)
+		} else if methodOption, ok := strictOrderingTypes[gorumsType]; ok {
+			genStrictOrderingMethods(data, methodOption)
```
meling, line 47: This can be replaced with:
```go
genGorumsMethods(data, methodOption)
```

cmd/protoc-gen-gorums/gengorums/gorums.go
meling: Line 69 can be replaced with:
```go
	genGorumsMethods(data, methodOption)
```

cmd/protoc-gen-gorums/gengorums/template_globals.go
```diff
@@ -0,0 +1,5 @@
+package gengorums
+
+var globals = `
```
meling, line 3: Suggest to move this to the `_types.go` file instead.

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -129,20 +176,40 @@ var gorumsTypes = map[string]*protoimpl.ExtensionInfo{
...
+}
+
+// mapping from strict ordering type to a checker that will check if a method has that type
+var strictOrderingTypeCheckers = map[*protoimpl.ExtensionInfo]func(*protogen.Method) bool{
+	strictordering.E_StrictOrderingQc: func(m *protogen.Method) bool {
+		return hasAllMethodOption(m, gorums.E_StrictOrdering, gorums.E_Qc)
+	},
+	strictordering.E_StrictOrderingRpc: func(m *protogen.Method) bool {
+		return hasMethodOption(m, gorums.E_StrictOrdering) && !hasMethodOption(m, gorumsCallTypes...)
+	},
```
meling, line 212: This is a neat way to organize the various checkers that we need; to that end, I'm wondering if we should add more checkers here or in a separate map, e.g.:
```go
	gorums.E_Multicast: func(m *protogen.Method) bool {
		return hasMethodOption(m, gorums.E_Multicast) && m.Desc.IsStreamingClient()
	},
... and so on ...
```
If we do something like this, maybe we can probably avoid some of the checks in the switch cases in `validateMethodExtensions`... 

Any thoughts?

cmd/protoc-gen-gorums/dev/strictordering.go
Raytar: I don't think it's necessary to record this error since it doesn't add any information about why the node stream was closed. Maybe we could return `node.LastErr()` instead of this generic error to give specific information about why the node errored? If `node.SetLastErr()` is called by `reconnectStream`, then it will probably contain the error that caused the stream to fail. Of course there is a chance that some unrelated error will have taken its place.

gorums.proto
Raytar: The service has now been moved to `dev`

cmd/protoc-gen-gorums/gengorums/template_strictordering_qc.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return nil, {{$errorf}}("failed to marshal message: %w", err)
+		}
+		msg := &GorumsMessage{
+			ID: msgID,
+			Method: "{{fullName .Method}}",
+			Data: data,
+		}
+		n.sendQ <- msg
+{{- else}}
```
meling, line 55: This else clause is not needed; just do the `n.sendQ <- msg` for both.

gorums.proto
```diff
@@ -14,6 +14,7 @@ extend google.protobuf.MethodOptions {
   bool qc_future = 50004;
   bool qf_with_req = 50005;
   bool per_node_arg = 50006;
+  bool strict_ordering = 50007;
```
meling, line 17: What if we used: `ordered` as the option name. Maybe this would also make sense elsewhere in the code, but maybe not everywhere.

cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -62,13 +63,23 @@ func GenerateFileContent(gen *protogen.Plugin, file *protogen.File, g *protogen.
...
+	genGorumsMethods(data, strictOrderingCallTypes...)
+	g.P()
 }
 
 func genGorumsMethods(data servicesData, methodOptions ...*protoimpl.ExtensionInfo) {
 	g := data.GenFile
 	for _, service := range data.Services {
 		for _, method := range service.Methods {
-			if hasMethodOption(method, methodOptions...) {
+			// skip all strict ordering methods.
```
meling, line 75: Remove these two comment lines, as they are no longer relevant.

PR #57, opened 2020-04-15
Move strictordering.proto into its own package

Raytar:
If `strictordering.proto` is bundled with the generated gorums code, then it will not be possible to have multiple services in the same codebase due to registration conflicts.
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict

This pull request moves `strictordering.proto` from `cmd/protoc-gen-gorums/dev` to its own package `strictordering`


Makefile
```diff
@@ -10,10 +10,11 @@ test_files				:= $(shell find $(tests_path) -name "*.proto" -not -path "*failing
 failing_test_files		:= $(shell find $(tests_path) -name "*.proto" -path "*failing*")
 test_gen_files			:= $(patsubst %.proto,%_gorums.pb.go,$(test_files))
 internal_so				:= internal/strictordering
+public_so				:= strictordering/
```
meling, line 13: Don't need the `/` at the end.

meling:
First, did you actually try to generate multiple services with Gorums? Even though much of the generator does support multiple services, we explicitly stop the generator if a proto file is supplied with multiple services defined, because there are other issues concerning this; don't recall the reasons right now, but we cannot have multiple services in one proto file. That said, I'm fine with promoting this to a separate package, as it may simplify future support for multiple services per proto file. On the other hand, it does expose the package as an "API" that others could use.

After thinking more about the naming (`strictordering`), I've come to the conclusion that this package and some of its other names should be changed. _Names are important, so let's get them right._ I've studied how the package name is used in our code, and there is a certain amount of stutter; something that the Go community tries to avoid.

Current names used in our code with this update:
- `strictordering.GorumsMessage`
- `strictordering.GorumsStrictOrdering_NodeStreamServer`
- `strictordering.RegisterGorumsStrictOrderingServer`
- `strictordering.GorumsStrictOrderingClient`
- `strictordering.GorumsStrictOrdering_NodeStreamClient`

This package actually refers to an _ordered stream connection_ between Gorums nodes (the servers). Here are some proposed alternatives to be discussed:

- File name: `ordering.proto`
- Package name: `ordering`. Other names considered: `connection` or `stream` or `transport` and moving back to root of repo. The latter approach, I'm not too happy with, since it exposes something that's really Gorums internal as if it is part of the API. 
- RPC Service name: just `Stream` or `Gorums`
- Message name: `Message` instead of `GorumsMessage`

These proposals would then give the following:
- `ordering.Message`
- `ordering.Stream_NodeStreamServer`
- `ordering.RegisterStreamServer`
- `ordering.StreamClient`
- `ordering.Stream_NodeStreamClient`

Or:
- `ordering.Message`
- `ordering.Gorums_NodeStreamServer`
- `ordering.RegisterGorumsServer`
- `ordering.GorumsClient`
- `ordering.Gorums_NodeStreamClient`

Raytar:
> First, did you actually try to generate multiple services with Gorums?

Yes, [hotstuff](https://github.com/relab/hotstuff) uses two unrelated services. [One](https://github.com/relab/hotstuff/blob/master/gorumshotstuff/internal/proto/hotstuff.proto) is used to implement the HotStuff algorithm itself, and [the other](https://github.com/relab/hotstuff/blob/master/clientapi/hotstuff.proto) is used in the client/server applications. In `hotstuffserver`, both of these packages are imported, and thus we get registration conflicts since both packages include the code from `strictordering.proto`.

I agree about the naming, and out of the two proposals, I prefer the first one.

meling:
Sounds good; as soon as you commit those changes, I'll merge.

Issue #56, opened 2020-04-14
Clean up Gorums method options


meling:
Currently, we have several options, some of which may be removed and others should be reorganized and renamed to make things easier to maintain.

- [x] We should rename the `gorums.qc` option to `gorums.quorumcall`. It is longer, but much clearer.
- [x] `gorums.qf_with_req` is often used, so perhaps we should once again reconsider whether or not we should require users to specify this; just make it the default.
- [x] `gorums.correctable_stream` can be replaced by `gorums.correctable` and the `stream` annotation at the output type (we can test for both of these in the code, so it shouldn't be difficult to implement). If necessary, we can follow a similar structure as the new `gorums.ordered` option.
- [x] Consider whether or not `gorums.qc_future` can be replaced with `gorums.correctable`. 
- [x] If we want to keep the future option, replace `gorums.qc_future` with two options: `gorums.quorumcall` and `gorums.future`. This can follow a similar structure as the new `gorums.ordered` option.

Before making hard decisions regarding the removal of certain options, e.g. future, we should implement test examples to use them to get more experience with it...

Opinions?

meling:
Some further thoughts. I've been thinking about the name `gorums.qc_future`. I think we should rename it to `gorums.async`, since that has become the de facto naming for such call semantics, even though it isn't exactly the async/await semantics of some languages. Thoughts? @tormoder @leandernikolaus @Raytar

Raytar:
> Some further thoughts. I've been thinking about the name `gorums.qc_future`. I think we should rename it to `gorums.async`, since that has become the de facto naming for such call semantics, even though it isn't exactly the async/await semantics of some languages. Thoughts? @tormoder @leandernikolaus @Raytar 

Is the idea still to make it work like the ordered option, such that you specify `gorums.quorumcall` and `gorums.async`? In that case, does it make sense to use it with any other call types? Can a `correctable` be async?

meling:
This is the question. I'm considering whether or not it makes sense to use `async` as a secondary option to `quorumcall`. Not sure it makes sense to use it along with the other call types, such as `correctable`, since correctable variants follow a similar design as that of `async`. To that end, maybe the correctable variants are really sub types of `async`. But there are essentially three variants:

- async (currently qc_future)
- correctable
- correctable_stream (currently this only requires correctable along with the rpc method being server streaming.)

tormoder:
(It's been a long time and I don't the bandwidth to look at the details).

[...]

PR #58, opened 2020-04-15
Add ordered qc_future

Raytar:
This pull request adds an ordered version of the `qc_future` calltype.
It is basically the same code as a regular ordered QC, but the code for receiving messages is run in a separate goroutine.

~I have marked it as a draft for now since it depends on #57.~
Now depends on #59


meling:
This must be adapted to the new generator design implemented in PR #59. Lots of stuff have changed, so expect it to take some time; maybe easier to close this and submit again, building on #59. 

Moreover, I'm still pondering whether or not to make the `future` calltype a subtype of the quorum call, but then this ordered `future` variant would need another nesting level. This has not been tested, but it should work. However, perhaps we should avoid this complexity. See also issue #56.

PR #59, opened 2020-05-04
reorganized the generator to handle sub call types

meling:
This PR allows to generate generic top-level call types with sub-behaviors; changing template for the nested call type. That is, we now support both `ordered` and `correctable` as top-level call types and nested types for `ordered_qc`, `ordered_rpc`, plain `correctable` and `correctable_stream`. This PR also tries to clean up some other stuff as well.


cmd/protoc-gen-gorums/gengorums/gorums_bundle.go
```diff
@@ -27,13 +28,13 @@ const (
 
 // GenerateBundleFile generates a file with static definitions for Gorums.
 func GenerateBundleFile(dst string) {
-	pkgIdent, code := bundle(devPkgPath)
+	pkgIdent, reservedIndents, code := bundle(devPkgPath)
```
Raytar, line 31: should it be `reservedIdents` instead of `reservedIndents`?

meling: fixed

Raytar:
I think it might be necessary to expand upon the `chkFn` pattern. If the `future` option is made a subtype of `quorumcall` and we merge #58, then the `chkFn` of the "unordered" future QC needs to check that the `ordered` option is not set. 

Instead of using `hasAllMethodOption` and `hasMethodOption`, we could write a `hasExactMethodOptions` that returns true only if the method's options exactly match the given options.

Another idea is to count how many options was matched by each `chkFn` and then choose the one that matched the most options. In this case we would have to make sure that none of the functions return the same number of matches for the same set of options.

meling:
Yes. This is correctly observed. It is a bit awkward as it is, but it can be solved by making sure that all table entries are exclusive, i.e. that it can't match on multiple entries. Basically, we need to negate some options, when checking for `quorumcall` for example. I haven't spent too much time finding the best solution yet, but I think doing something along the lines of exact matching would be the way to go... I was thinking to replace the `chkFn` pattern with a list of options to match against, but it isn't flexible enough to also capture the need to match with `stream` method checks.

PR #62, opened 2020-05-19
Optimize messages of ordered calltypes

Raytar:
This PR reduces the size of the messages in all ordered calltypes by removing unnecessary URLs that were embedded in the messages.

Previously, we embedded the method URL within the `NodeStream` `Message` such that the server-side code is able to determine which handler to execute. We can instead replace this with a unique numeric ID for each method.

We also used to use the protobuf `Any` format to encode the request and response messages. The `Any` format embeds the URL of the message type. We can instead marshal the messages directly to a `[]byte`.

I used Wireshark to inspect the message size of the `ReadResponse` message (shown below) when sent within an ordered QC.

```proto
message State {
	string Data = 1;
	uint64 Timestamp = 2;
}

message ReadResponse {
	bool OK = 1;
	State State = 2;
}
```

Before this change, the packet size was 163 bytes and the gRPC message size was 83 bytes. With this change, the packet size was 93 bytes and the gRPC message size was 13 bytes.

Before: 
```
0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
0010   00 95 3d a4 40 00 40 06 fe bc 7f 00 00 01 7f 00   ..=.@.@.........
0020   00 01 a2 94 15 b3 89 f4 ca 30 4b 62 c0 50 80 18   .........0Kb.P..
0030   02 00 fe 89 00 00 01 01 08 0a 64 62 2a 19 64 62   ..........db*.db
0040   22 6c 00 00 58 00 00 00 00 00 01 00 00 00 00 53   "l..X..........S
0050   08 02 12 1d 2f 70 72 6f 74 6f 2e 47 6f 72 75 6d   ..../proto.Gorum
0060   73 54 65 73 74 2f 52 65 61 64 4f 72 64 65 72 65   sTest/ReadOrdere
0070   64 1a 30 0a 25 74 79 70 65 2e 67 6f 6f 67 6c 65   d.0.%type.google
0080   61 70 69 73 2e 63 6f 6d 2f 70 72 6f 74 6f 2e 52   apis.com/proto.R
0090   65 61 64 52 65 71 75 65 73 74 12 07 0a 05 68 65   eadRequest....he
00a0   6c 6c 6f                                          llo
```

After:
```
0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
0010   00 4f 04 99 40 00 40 06 38 0e 7f 00 00 01 7f 00   .O..@.@.8.......
0020   00 01 a4 2e 15 b3 90 33 58 8b ac ce f5 1d 80 18   .......3X.......
0030   02 00 fe 43 00 00 01 01 08 0a 64 72 bb b6 64 72   ...C......dr..dr
0040   b5 d1 00 00 12 00 00 00 00 00 01 00 00 00 00 0d   ................
0050   08 02 10 01 1a 07 0a 05 68 65 6c 6c 6f            ........hello
```


ordering/ordering.proto
```diff
@@ -3,12 +3,10 @@ syntax = "proto3";
...
 
-import "google/protobuf/any.proto";
-
 service Gorums { rpc NodeStream(stream Message) returns (stream Message); }
 
 message Message {
   uint64 ID = 1;
-  string Method = 2;
-  google.protobuf.Any Data = 3;
+  uint32 MethodID = 2;
```
meling, line 10: Why not make this `int32` to avoid all the type conversion noise.

Raytar: Yeah, I'll change them all to the same type.

cmd/protoc-gen-gorums/gengorums/template_datatypes.go
```diff
@@ -1,9 +1,30 @@
...
+)
+
 var globals = `
 const hasOrderingMethods = {{hasOrderingMethods .Services}}
 `
 
+var orderingInit = `
+{{- range .Services }}
+{{- range orderedMethods .Methods }}
+var {{.GoName}}MethodID int
```
meling, line 15: I think we can easily make these into `const` instead of variables. Then we don't need the `init()` function, since we can initialize them at generation time.

Raytar: How would you do that? Using a function in `func-map`?

Raytar: I guess I could do something like this:

 `{{- $index, $element := range orderedMethods .Method }}`

 and use `$index` as the ID.

meling: Yes, that looks like a reasonable way to do it. Also, do these consts need to be exported?

meling: Ah, ok. I see they are just copying the method names and appending MethodID.

meling: Perhaps we should use the unexport func on these, to avoid making them exported (exposing them as API)...

Raytar: No, I don't think so. 

By the way, does the `{{- range .Services }}` ever run more than once?

meling: No not now, since Gorums currently doesn't support more than one service per proto file. But that's a restriction that could go away in a future upgrade... So, I favor keeping the iterations over all services everywhere we can, so that upgrading it becomes easier in the future...

cmd/protoc-gen-gorums/dev/ordering.go
```diff
@@ -13,12 +13,18 @@ import (
 	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"google.golang.org/protobuf/types/known/anypb"
 )
 
+var numOrderedMethods int32 = 0
```
meling, line 18: This and the `getOrderedMethodID()` below is no longer needed.

Raytar: Oops, forgot about those

PR #66, opened 2020-05-27
Adjustments to ordered quorum calls

Raytar:
This PR makes some adjustments to the API of the ordered quorum call variants.

For the regular ordered quorum call, the `grpc.CallOption` field can be removed, as it is not applicable to ordered quorum calls.

For the async ordered quorum call, the quorum call method no longer returns an error along with the future object. Instead, the error is returned by the `Get()` method on the future object.

In addition, I have refactored the template for the async ordered quorum call such that the additional "send" function is not necessary. Instead, the sending is done directly in the quorum call method.


cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
```diff
@@ -529,48 +529,51 @@ func (s *GorumsServer) RegisterOrderingUnaryRPCHandler(handler OrderingUnaryRPCH
...
-
-func (c *Configuration) orderingFutureSend(ctx context.Context, in *Request) (uint64, int, chan *orderingResult, error) { // get the ID which will be used to return the correct responses for a request
+	// get the ID which will be used to return the correct responses for a request
 	msgID := c.mgr.nextMsgID()
 
 	// set up a channel to collect replies
 	replyChan := make(chan *orderingResult, c.n)
 	c.mgr.putChan(msgID, replyChan)
 
+	expected := c.n
```
meling, line 544: We can probably avoid the `expected` variable and just use `c.n` or `c.Size()`; at least I don't see any changes to expected in the code.

cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
```diff
@@ -529,48 +529,51 @@ func (s *GorumsServer) RegisterOrderingUnaryRPCHandler(handler OrderingUnaryRPCH
...
+
+	var msg *ordering.Message
 	data, err := proto.MarshalOptions{AllowPartial: true, Deterministic: true}.Marshal(in)
 	if err != nil {
-		return 0, 0, nil, fmt.Errorf("failed to marshal message: %w", err)
+		// In case of a marshalling error, we should skip sending any messages
```
meling, line 549: `goto End` will actually try to send stuff...

meling: Okay, maybe not... But it was a bit confusing to not return here.

cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
Raytar: It only changes if `perNodeArg` is enabled.

cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
Raytar: I agree, but the goroutine for the `Recv` method must be started before returning.
The other ways that I can think of to achieve the same result would be to either
start the goroutine before returning in both places:

```go
if err != nil {
	// In case of a marshalling error, we should skip sending any messages
	fut.err = fmt.Errorf("failed to marshal message: %w", err)
	go c.{{$unexportMethod}}Recv(ctx, in, msgID, expected, replyChan, fut)
	return fut
}
```

or to `defer` it like this (`defer go` is not supported):

```go
defer func() { go c.{{$unexportMethod}}Recv(ctx, in, msgID, expected, replyChan, fut) }()
```

meling: Actually, the only thing that happens in this case, after starting the goroutine is:

```go
	defer close(fut.c)

 	if fut.err != nil {
 		return
 	}
```

Hence, we can do this instead:

```go
if err != nil {
	// In case of a marshalling error, we should skip sending any messages
	fut.err = fmt.Errorf("failed to marshal message: %w", err)
	close(fut.c)
	return fut
}
```

I think this is more clear than actually having to start the goroutine to close the channel, even though we repeat the close operation in two places then...

Another alternative would be to not initialize the `fut.c` channel, until it is needed inside the `orderingFutureRecv` goroutine.

Raytar: Nice, I didn't think of that. I will go ahead and implement it. Since the goroutine is not started in case of an error, there should be no problem with closing the channel there.

PR #70, opened 2020-06-11
Add an option to control the size of sendQ buffer

Raytar:
In calltypes using `NodeStream`, the size of the buffer in the `sendQ` channel can have a major impact on performance for some calltypes. In particular, a small buffer yields lower latency, while a larger buffer yields higher throughput.

This PR adds a new `ManagerOption`, `WithSendBufferSize`, that gives control over the size of the buffer on this channel.

I have done some benchmarks comparing the new implementation of Multicast (#65) with the old version, using different buffer sizes. The benchmarks were performed on some of the `pitter` machines at UiS, with 1 client and 4 servers.

throughput
latency

Raw data:

Baseline (old multicast):

```
Benchmark    Througput           Latency    Std.dev    Client+Servers                    
Multicast    86596.45 ops/sec    0.44 ms    0.82 ms    3101 B/op         61 allocs/op 
```

New Multicast:

```
Buffer size: 0
Benchmark    Througput           Latency    Std.dev    Client+Servers                    
Multicast    55255.25 ops/sec    0.27 ms    0.51 ms    4399 B/op         83 allocs/op    

Buffer size: 1
Benchmark    Througput           Latency    Std.dev    Client+Servers                    
Multicast    62638.19 ops/sec    0.31 ms    0.54 ms    4464 B/op         83 allocs/op    

Buffer size: 10
Benchmark    Througput            Latency    Std.dev    Client+Servers                    
Multicast    100948.09 ops/sec    0.50 ms    0.89 ms    4529 B/op         83 allocs/op    

Buffer size: 100
Benchmark    Througput            Latency    Std.dev    Client+Servers                    
Multicast    156563.76 ops/sec    1.67 ms    1.85 ms    4547 B/op         83 allocs/op    

Buffer size: 1000
Benchmark    Througput            Latency    Std.dev     Client+Servers                    
Multicast    153235.60 ops/sec    9.49 ms    17.98 ms    4652 B/op         83 allocs/op    
```

### DIFF
--- a/cmd/protoc-gen-gorums/dev/mgr.go
+++ b/cmd/protoc-gen-gorums/dev/mgr.go
@@ -43,9 +43,7 @@ func NewManager(nodeAddrs []string, opts ...ManagerOption) (*Manager, error) {
 		lookup:       make(map[uint32]*Node),
 		configs:      make(map[uint32]*Configuration),
 		receiveQueue: newReceiveQueue(),
-		opts: managerOptions{
-			backoff: backoff.DefaultConfig,
-		},
+		opts:         newManagerOptions(),
 	}
 
 	for _, opt := range opts {
@@ -110,7 +108,7 @@ func (m *Manager) createNode(addr string) (*Node, error) {
 		addr:    tcpAddr.String(),
 		latency: -1 * time.Second,
 	}
-	node.createOrderedStream(m.receiveQueue, m.opts.backoff)
+	node.createOrderedStream(m.receiveQueue, m.opts)
 
 	return node, nil
 }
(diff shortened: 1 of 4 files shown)


Raytar:
Also note that I had to modify the "old" multicast in order to make it work properly. In particular, I removed the extra goroutine that is started per node for each multicast call:

```go
func (c *Configuration) {{$method}}(in *{{$in}}) error {
	for _, node := range c.nodes {
		err := node.{{unexport $method}}Client.Send(in)
		if err == nil {
			continue
		}
		if c.mgr.logger != nil {
			c.mgr.logger.Printf("%d: {{$method}} stream send error: %v", node.id, err)
		}
	}
	return nil
}
```
Without this tweak, it becomes impossible to know whether a multicast has been sent. Thus, if executing a multicast call in a loop, the system can run out of memory because it creates new goroutines to send messages faster than messages are sent.

PR #71, opened 2020-06-15
Generate a server interface for NodeStream methods

Raytar:
Previously, each NodeStream method required its handler to be registered with the server using a `Register{...}Handler` function. With this change, all of the handlers for a service can be registered at once by calling a `Register{...}Server` function, similarly to the way it is done in gRPC.


cmd/protoc-gen-gorums/dev/zorums_types_gorums.pb.go
```diff
@@ -590,3 +593,215 @@ func (c *CorrectableStreamResponse) set(reply *Response, level int, err error, d
...
+	OrderingFutureCustomReturnType(*Request) *Response
+	OrderingFutureCombo(*Request) *Response
+	Unicast(*Request)
+	Unicast2(*Request)
+}
+
+func (s *GorumsServer) RegisterZorumsServiceServer(srv ZorumsService) {
+	s.srv.handlers[multicastMethodID] = func(in *ordering.Message) *ordering.Message {
+		req := new(Request)
+		err := proto.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}.Unmarshal(in.GetData(), req)
```
meling, line 620: Since these unmarshal and marshal options (below) are the same for all, would we save some allocations, if we defined it once, and reused the same in all handlers?

cmd/protoc-gen-gorums/dev/zorums_types_gorums.pb.go
```diff
@@ -590,3 +593,215 @@ func (c *CorrectableStreamResponse) set(reply *Response, level int, err error, d
...
+		if err != nil {
+			return nil
```
meling, line 622: Should we add `error` to the interface methods, to allow these errors to be reported?

cmd/protoc-gen-gorums/dev/zorums_types_gorums.pb.go
```diff
@@ -590,3 +593,215 @@ func (c *CorrectableStreamResponse) set(reply *Response, level int, err error, d
...
+		}
+		srv.Multicast4(req)
+		return nil
+	}
+	s.srv.handlers[orderingQCMethodID] = func(in *ordering.Message) *ordering.Message {
+		req := new(Request)
+		err := proto.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}.Unmarshal(in.GetData(), req)
+		// TODO: how to handle marshaling errors here
```
meling, line 666: Should we add `error` to the interface methods, to allow these errors to be reported?

cmd/protoc-gen-gorums/dev/zorums_types_gorums.pb.go
Raytar: It would indeed be nice to be able to report these errors, but I have not yet thought of a good way to do that from the `NodeStream` handler.

PR #72, opened 2020-06-15
Add a `concurrent` option

Raytar:
This PR adds a `concurrent` option, which makes the server-side handlers of `NodeStream` methods run concurrently.

The server runs multiple goroutines to accomplish this. One goroutine receives messages and forwards them to other goroutines. One goroutine is dedicated to all of the `ordered` messages. Another goroutine is responsible for sending replies. All methods that use the `concurrent` option get executed in a new goroutine.

I have added a server option `WithServerBufferSize` that controls the buffer on some of the channels that are used. This should allow for some performance tuning, similarly to #70.


cmd/protoc-gen-gorums/gengorums/gorums_func_map.go
```diff
@@ -105,6 +105,9 @@ var funcMap = template.FuncMap{
 	"isOneway": func(method *protogen.Method) bool {
 		return hasMethodOption(method, gorums.E_Multicast, gorums.E_Unicast)
 	},
+	"hasConcurrentHandler": func(method *protogen.Method) bool {
```
meling, line 108: Maybe `isConcurrent` is a good name for this?

cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
```diff
@@ -460,10 +460,120 @@ func (s *GorumsServer) RegisterOrderingComboHandler(handler OrderingComboHandler
...
+	msgID := c.mgr.nextMsgID()
+
+	// set up a channel to collect replies
+	replies := make(chan *orderingResult, c.n)
+	c.mgr.putChan(msgID, replies)
+
+	// remove the replies channel when we are done
+	defer c.mgr.deleteChan(msgID)
+
+	data, err := proto.MarshalOptions{AllowPartial: true, Deterministic: true}.Marshal(in)
```
meling, line 500: Same for these marshal options (and unmarshal) below. Maybe these could be using some globally defined option variable?

PR #73, opened 2020-06-16
fix linter warnings about deprecated fields

Raytar:
Accessing the fields of the ExtensionInfo structs from the generated
protocol buffers is deprecated.

This commit uses the TypeDescriptor embedded in the ExtensionInfo
instead, as this is part of the new API.

### DIFF
--- a/cmd/protoc-gen-gorums/gengorums/gorums.go
+++ b/cmd/protoc-gen-gorums/gengorums/gorums.go
@@ -192,6 +192,12 @@ func (c *callTypeInfo) deriveCallType(m *protogen.Method) *callTypeInfo {
 	return c
 }
 
+// callTypeName returns the name of the call type option with the prefix removed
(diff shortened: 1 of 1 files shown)


cmd/protoc-gen-gorums/gengorums/gorums.go
```diff
@@ -192,6 +193,12 @@ func (c *callTypeInfo) deriveCallType(m *protogen.Method) *callTypeInfo {
 	return c
 }
 
+// callTypeName returns the name of the call type option with the prefix removed
+func callTypeName(ext protoreflect.ExtensionDescriptor, prefixLen int) string {
```
meling, line 197: If you make this take `*protoimpl.ExtensionInfo`, you can do the call to `TypeDescriptor()` her instead.

Raytar: Cool, I didn't realize that there was a non-internal alias to that type.

PR #80, opened 2020-06-24
Improve marshaling/unmarshaling performance for NodeStream calltypes

### LINKED ISSUE #78: The ordered QC implementation performs worse than the old QC implementation with larger messages
With small messages, the ordered QC implementation is faster because it reuses the same stream for sending messages, whereas the old (unary RPC based) QC implementation has to allocate memory for a new `ClientStream` on every call. However, when the message size becomes large enough, the memory required to create the streams becomes insignificant. When this happens, the old QC implementation outperforms the new one because it only has to marshal a single message. In the new implementation, the application message must first be marshaled before the `ordering.Message` can be marshaled. This means that the payload gets copied twice.

On my machine, the old QC implementation starts performing better than the ordered QC implementation at a message size of around 45-50KB.

To improve this performance, we need to prevent the application message from being copied several times. I can see a few different ways to solve this:

* Write a gRPC [Codec](https://pkg.go.dev/google.golang.org/grpc/encoding?tab=doc#Codec) to manually marshal the messages. Manually marshaling the messages should be doable using the [protowire](https://pkg.go.dev/google.golang.org/protobuf/encoding/protowire?tab=doc) package. The [proto](https://pkg.go.dev/google.golang.org/protobuf/proto?tab=doc) package contains functions that would make this a lot easier, but most of them are unexported.

* Marshal both messages separately onto the same buffer and send them. This way, the application message does not have to be embedded in the `ordering.Message`. This can already be done with the [`MarshalAppend`](https://pkg.go.dev/google.golang.org/protobuf/proto?tab=doc#MarshalOptions.MarshalAppend) function. ~However, I could not find an easy way to unmarshal the messages.~ Unmarshaling can be done as long as the messages are prefixed with their size.

* Somehow generate the `ordering.Message` such that it includes the application messages as fields. We could use the `oneof` type for this. We would need to make a template for the `ordering.proto` file, output it from the gorums plugin and then compile it. This would probably be the best solution if it can be implemented seamlessly.
  * Problem: Difficult / impossible to determine the correct import paths to generate the proto file without requiring additional plugin options or requiring the user to compile the generated proto file manually.

* Embed the method and message IDs in the application message's `UnknownFields`. The problem with this solution is that the application message must be unmarshaled before the metadata can be accessed, which means that the type of the message must be known before the metadata is available. Implementing this would likely require each method to have its own gRPC stream.

* Ask if this functionality could be added to protobuf.

Raytar:
This PR implements a gRPC [`Codec`](https://pkg.go.dev/google.golang.org/grpc/encoding?tab=doc#Codec) that performs fast marshaling and unmarshaling of a gorums message consisting of metadata and an application-specific data message. This is accomplished by marshaling both the metadata and data messages onto the same byte slice using [`MarshalAppend`](https://pkg.go.dev/google.golang.org/protobuf/proto?tab=doc#MarshalOptions.MarshalAppend) and prefixing each marshaled message with its length. Unmarshaling is simply done by reading the length of each message and unmarshaling them.

This approach offers better performance than the old approach because it avoids copying the data message more than once. See #78 for more details.

Fixes #78


Raytar:
Benchmark of this new approach vs the old:
```
goos: linux
goarch: amd64
pkg: github.com/relab/gorums/tests/message_encoding
BenchmarkMetadataWithEmbeddedMessage/Marshal/10-8        1000000              2235 ns/op             240 B/op          5 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Marshal/100-8        415006              2829 ns/op             432 B/op          5 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Marshal/1000-8       337478              4005 ns/op            2224 B/op          5 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Marshal/10000-8       65666             15400 ns/op           20656 B/op          5 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Marshal/100000-8       8941            133444 ns/op          213168 B/op          5 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Unmarshal/10-8       369932              2779 ns/op             240 B/op          6 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Unmarshal/100-8      448782              2891 ns/op             432 B/op          6 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Unmarshal/1000-8     251708              4706 ns/op            2240 B/op          6 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Unmarshal/10000-8     64753             16017 ns/op           20672 B/op          6 allocs/op
BenchmarkMetadataWithEmbeddedMessage/Unmarshal/100000-8     9732            122785 ns/op          213184 B/op          6 allocs/op
BenchmarkGorumsCodec/Marshal/10-8                         609368              1832 ns/op             184 B/op          4 allocs/op
BenchmarkGorumsCodec/Marshal/100-8                        588836              2003 ns/op             280 B/op          4 allocs/op
BenchmarkGorumsCodec/Marshal/1000-8                       445206              2631 ns/op            1176 B/op          4 allocs/op
BenchmarkGorumsCodec/Marshal/10000-8                      125390              8403 ns/op           10392 B/op          4 allocs/op
BenchmarkGorumsCodec/Marshal/100000-8                      17895             65130 ns/op          106648 B/op          4 allocs/op
BenchmarkGorumsCodec/Unmarshal/10-8                       630361              1966 ns/op             192 B/op          5 allocs/op
BenchmarkGorumsCodec/Unmarshal/100-8                      593460              2616 ns/op             288 B/op          5 allocs/op
BenchmarkGorumsCodec/Unmarshal/1000-8                     363512              3448 ns/op            1200 B/op          5 allocs/op
BenchmarkGorumsCodec/Unmarshal/10000-8                    108781              9518 ns/op           10416 B/op          5 allocs/op
BenchmarkGorumsCodec/Unmarshal/100000-8                    19300             59640 ns/op          106672 B/op          5 allocs/op
```

Raytar:
Quorum call benchmarks of the new vs the old:

```
New 0 payload:

Benchmark      Throughput         Latency    Std.dev    Client+Servers                     
OrderedQC      3901.46 ops/sec    0.26 ms    0.26 ms    1647 B/op         54 allocs/op     
UnorderedQC    2243.68 ops/sec    0.44 ms    0.45 ms    8743 B/op         168 allocs/op    

Old 0 payload: 
                    
OrderedQC      3915.95 ops/sec    0.25 ms    0.26 ms    1607 B/op         49 allocs/op     
UnorderedQC    2245.67 ops/sec    0.44 ms    0.48 ms    8683 B/op         166 allocs/op    

New 1M payload:  
               
OrderedQC      149.72 ops/sec    6.68 ms    0.82 ms    6199415 B/op      225 allocs/op    
UnorderedQC    142.94 ops/sec    6.99 ms    1.00 ms    6215872 B/op      353 allocs/op    

Old 1M payload:     
            
OrderedQC      118.64 ops/sec    8.43 ms    2.68 ms    10477961 B/op     240 allocs/op    
UnorderedQC    139.99 ops/sec    7.14 ms    1.47 ms    6213726 B/op      352 allocs/op    
```

cmd/protoc-gen-gorums/gengorums/template_datatypes.go
```diff
@@ -172,31 +173,19 @@ type {{$service}} interface {
...
+	s.srv.handlers[{{unexport .GoName}}MethodID] = func(in *gorumsMessage) *gorumsMessage {
+		req := in.message.(*{{in $genFile .}})
 		{{- if isOneway .}}
-		if err != nil {
-			return nil
-		}
 		srv.{{.GoName}}(req)
 		return nil
 		{{- else}}
 		// TODO: how to handle marshaling errors here
```
meling, line 186: Looks like this is no longer needed then?

ordering/ordering.proto
```diff
@@ -5,8 +5,9 @@ option go_package = "github.com/relab/gorums/ordering";
 
 service Gorums { rpc NodeStream(stream Message) returns (stream Message); }
 
-message Message {
-  uint64 ID = 1;
+message Message {}
```
Raytar, line 8: I think I might remove this "placeholder" message and use `Metadata` instead for the `NodeStream`, even though the type that is actually being sent is `gorumsMessage` which contains `Metadata`.

meling: Right, I was wondering (but didn't check carefully); do we still need the exported `ordering` package? I see several places `ordering.Message` gets replaced with `gorumsMessage`. I guess this comes back to the question, can we move `NodeStream` and the `Metadata` message into `dev`? Maybe not.

Raytar: We still need the `Metadata` message, unless we want to marshal that manually as well. That means that we can't bundle it, as that will lead to registration conflicts if multiple gorums services are imported in the same project.

PR #81, opened 2020-06-24
Make it possible for server-side handlers to run asynchronously

Raytar:
This PR reworks server-side handlers such that they are able to run asynchronously. This is done by passing a channel to the handler. Instead of returning a response message, the handler sends it on the channel. This allows users to launch goroutines within the handlers to do async tasks. For example:

```go
func (s *testSrv) AsyncHandler(req *Request, c chan<- *Response) {
	response := &Response{
		InOrder: s.isInOrder(req.GetNum()),
	}
	go func() {
		// simulate some async action
		time.Sleep(10 * time.Millisecond)
		c <- response
	}()
}
```

In this example, we can also see that this pattern allows some parts of the handler to run synchronously, while slower or asynchronous tasks can be performed in a goroutine.

Because these handlers can launch goroutines to perform operations concurrently, it is no longer necessary to have a separate `concurrent` option.

Performance-wise, this new handler pattern is slower for fully synchronous handlers (seems to be a 3% performance hit worst-case). This is because a goroutine is started to receive the response message from each handler, wrap it in a `gorumsMessage`, and send it on the `finished` channel. It may be possible to avoid these goroutines and send the messages directly on the `finished` channel.


tests/ordering/order_test.go
```diff
@@ -241,6 +256,55 @@ func TestQCFutureOrdering(t *testing.T) {
...
+	}
+	c, err := man.NewConfiguration(man.NodeIDs(), &testQSpec{})
+	if err != nil {
+		t.Fatalf("Failed to create configuration: %v", err)
+	}
+
+	// begin test
+	numRuns := 1 << 16
+	for i := 1; i < numRuns; i++ {
+		resp, err := c.QC(context.Background(), &Request{Num: uint64(i)})
```
meling, line 293: Doesn't seem to be calling the `AsyncHandler()` RPC.

Raytar: I should probably think while using copy-paste :facepalm:

tests/ordering/order_test.go
```diff
@@ -84,35 +88,49 @@ func (q testQSpec) QCFutureQF(_ *Request, replies []*Response) (*Response, bool)
...
 }
 
-func TestUnaryRPCOrdering(t *testing.T) {
-	defer leakcheck.Check(t)
-	// server setup
-	tSrv := &testSrv{}
-	srv := NewGorumsServer()
-	srv.RegisterGorumsTestServer(tSrv)
-	lis, err := getListener()
+func setup(t *testing.T, cfgSize int) (cfg *Configuration, teardown func()) {
```
meling, line 95: This seems to overlap with `TestSetup` in `testing_gorums.go`. Perhaps we can make adjustments there and reuse it.

Raytar: I can update `testing_gorums.go` to use `GorumsServer` instead of `grpc.Server` and then we can use that

doc/userguide.md
```diff
@@ -138,22 +138,23 @@ type storageSrv struct {
   state *State
 }
 
-func (srv *storageSrv) Read(req *ReadRequest) *State {
+func (srv *storageSrv) Read(req *ReadRequest, out chan<- *State) {
```
meling, line 141: @leandernikolaus What do you think about this API change?? It will impact compatibility with previous Raft and other Gorums implementations.

tests/ordering/order_test.go
Raytar: Actually, this is not as easy as I thought. `GorumsServer` is part of the generated code, so it is difficult to make it work with a function like `TestSetup`.

meling: That's why `TestSetup()` takes a `regSrvFn`, but not sure if `GorumsServer` changes its interface

Raytar: `regSrvFn` should work the same, but we also need to create new instances of `GorumsServer`.

cmd/protoc-gen-gorums/dev/server.go
```diff
@@ -61,33 +47,14 @@ func (s *orderingServer) NodeStream(srv ordering.Gorums_NodeStreamServer) error
...
-			case <-ctx.Done():
-				return
-			case req := <-ordered:
-				handleMsg(req.msg, req.info)
-			}
-		}
-	}()
-
 	for {
 		req := newGorumsMessage(false)
```
meling, line 51: Shouldn't this be `true`?

cmd/protoc-gen-gorums/dev/server.go
```diff
@@ -61,33 +47,14 @@ func (s *orderingServer) NodeStream(srv ordering.Gorums_NodeStreamServer) error
 		}
```
meling, line 47: The go func() above may exit if `srv.SendMsg()` return an `err`. Would this error message be relevant to return from the `NodeStream` func. Or alternatively, it would be good to have a comment that explain why it's okay to exit this goroutine without exiting from the `NodeStream` func.

Raytar: If either `SendMsg()` or `RecvMsg()` return an `error` then the stream is aborted. So, if an error happens in the sending goroutine, then the receiving goroutine should also get an error and vice-versa.

cmd/protoc-gen-gorums/dev/server.go
Raytar: The `reply` field in the `gorumsMessage` type is used to tell the `gorumsUnmarshal` function whether it should create a new instance of the request type or the response type. On the server, we want to unmarshal the Request type, and hence it is set to `false`. I suppose I could create some constants to use in place of `true/false`. Might make it easier to read.

Raytar: Or perhaps it could be an argument to the `gorumsUnmarshal` function instead...

meling: It was just confusing when `reply=false` was used to mark a response message. You could  also just change the name of the input to `request` instead of `reply`. Consts could also work.  Didn’t look at the last suggestion in depth. So maybe that’s better...

Raytar: I have added the consts `gorumsRequest` and `gorumsResponse`, which are used instead of a boolean.

doc/userguide.md
leandernikolaus: I don't mind breaking changes.

However, I'm not totally convinced that the benefits in this change outweigh the added complexity, especially for a first time user.

Raytar: These are the pros/cons that I can see:

Pros:
  * User gets control over handler's concurrency
  * Handlers can be started in the correct order, but do expensive work concurrently
  * The same API can be used for streaming calls (correctable?)
  * Fewer options in the Gorums plugin / simplifies gorums code

Cons:
  * Requires changes to existing projects
  * API is no longer similar to gRPC
  * Slight performance penalty due to starting more goroutines.

Raytar:
@meling I configured the cSpell VSCode plugin to make it easier to work with in Markdown, but I can remove that commit if you don't like it.

Raytar:
Some additional thoughts about this API and gRPC:

Within gRPC, an API like this is not necessary or beneficial. This is because gRPC invokes each server handler in its own goroutine (see the [grpc-go documentation][grpc-concurrency]), and they don't care about ordering between separate RPCs. Handlers can run for as long as they need without impacting other RPCs.

In Gorums, we want to preserve the ordering of messages between a client and a single node. That means that request handlers must be executed synchronously by the server. The problem is that the execution of the handlers may slow down the processing of requests since it all happens on a single thread. However, the application may only need to worry about ordering up to a certain point. For example, in HotStuff, we worry about the ordering of requests until the requests are placed in a queue. After that, the handler needs to wait an unknown amount of time until the request has been executed by the protocol before returning a response. This is easily implemented using this API. The handler must simply add the request to the queue, and then it can start a goroutine that waits for HotStuff to execute the request. When the request has been executed, the response can be sent back via the `out` channel. That way the penalty for running server handlers synchronously is reduced somewhat, and ordering can be preserved.

[grpc-concurrency]: https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md

PR #88, opened 2020-07-07
More server API updates

Raytar:
This PR changes the server API again...

Instead of using channels to send responses, we can use a simple `func` that embeds the message in metadata and forwards onto the correct channel. Thus, we no longer have to create a new goroutine for every single message.

This PR also adds the server stream context as a parameter to each handler, as well as new manager options for setting metadata. The metadata can be extracted from the stream context:

```go
metadata.FromIncomingContext(ctx)
```

In addition, I have found a way to make the `TestSetup` helper useful for `GorumsServer` types.

The new server API:

```go
func (srv *orderedServer) OrderedQC(_ context.Context, in *Echo, out func(*Echo)) {
	out(in)
}
```

~Am leaving this as a draft for now, as one of the tests is failing. Will investigate tomorrow...~

EDIT: failing ordering test was due to indexing QF map incorrectly.


Raytar:
@meling I added a new test for TLS. I think that the old test broke because the included certificate expired. To avoid that problem, the new test generates its own TLS certificate.

Raytar:
@meling I've also added a test that verifies that reconnection of the streams is working. Let me know if you want me to break some of these changes out into separate pull requests.

Raytar:
Also, should we consider not including all of the generated code for the tests, and rely on the Makefile instead? The amount of code that gets updated from a single change in gorums is quite large now. We could set up a Makefile target to run all of the tests and then set up a GitHub action to run the tests automatically.

cmd/protoc-gen-gorums/gengorums/template_server.go
```diff
@@ -23,22 +27,23 @@ var registerInterface = `
...
-		srv.{{.GoName}}(req)
+		srv.{{.GoName}}(ctx, req)
 		{{- else }}
-		c := make(chan *{{out $genFile .}})
-		go func() {
-			resp := <-c
-			finished <- &gorumsMessage{metadata: in.metadata, message: resp}
-		}()
-		srv.{{.GoName}}(req, c)
+		once := new({{use "sync.Once" $genFile}})
```
meling, line 35: Could you add a brief comment here explaining why we use the `once.Do()` thing?

Also, does the `finished` channel mean the queue of messages to be sent on the stream? I think here it sounds okay to use `finished` (maybe), but in the `NodeStream` implementation, it is also called finished. Maybe this should be called something else? (outQueue and sendQueue?)

meling:
> Also, should we consider not including all of the generated code for the tests, and rely on the Makefile instead? The amount of code that gets updated from a single change in gorums is quite large now. We could set up a Makefile target to run all of the tests and then set up a GitHub action to run the tests automatically.

@Raytar Agree with this. I've been wanting to set up CI testing, as mentioned in #90. We should still keep the zorums* generated files in dev, since they are required to compile the static files.

meling:
Thanks for adding tests and the TLS cert generation. Maybe it will be easier to keep multiple PRs, if we don't check in all the generated code. I guess we will still get conflicts in the zorums* files.

Raytar:
Closing in favor of #91 and #92.

PR #93, opened 2020-07-10
New Gorums Stability Checker

meling:
This PR replaces the Makefile-based stability checker with TestGorumsStability,
which should be easier to run as part of a CI system, and in general clean up
some useless *.pb.go files.

- a few more words in the spell check dict
- removed unused funcs in gorums_test.go
- frist draft version of stability test
- Make test target depend on installgorums
- removed stability test files no longer used
- finalized stability checker in Go, obviating the need for the make target


meling:
@Raytar can you also take a look at why the golang-ci thing is failing?

Raytar:
> @Raytar can you also take a look at why the golang-ci thing is failing?

It's probably because the generated test protos were deleted. We will likely have to add a `make` step to the golangci workflow to compile those before linting.

meling:
Right. Maybe golangci-lint can be done in the same workflow as the others you have added to avoid having to do this make step twice!?

PR #98, opened 2020-07-20
Remove the `grpc.CallOption` field from ordered RPCs

Raytar:
This PR removes the `opts ...grpc.CallOption` field from Ordered Unary RPCs. The field was not actually used for anything and I probably forgot to remove it after copying some code... The field is variadic, so it should not break any existing code as long as it did not attempt to use any call options (which would have no effect anyway).

### DIFF
--- a/cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
+++ b/cmd/protoc-gen-gorums/dev/zorums_ordered_gorums.pb.go
@@ -6,7 +6,6 @@ import (
 	context "context"
 	ordering "github.com/relab/gorums/ordering"
 	trace "golang.org/x/net/trace"
-	grpc "google.golang.org/grpc"
 	time "time"
 )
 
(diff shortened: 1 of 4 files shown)


meling:
Are you sure this won't be useful in the future. I added these call options in other call types for compatibility with gRPC, even though I currently don't use them for anything...

Raytar:
Sure, it might be useful to have something like this in the future, but I can't think of a way to apply these options to an existing stream. These options are meant for gRPC's RPCs. If we wanted something similar, we would likely need to implement it ourselves. Because we use a single stream per node, the only time that gRPC call options can be specified in Gorums is when the stream is created. This can be done using `grpc.WithDefaultCallOptions` I believe.

meling:
Maybe it is better to add a comment about this to each Ordered Unary that the call option is currently not supported, but may be implemented in the future??

Raytar:
I don't see much value in keeping it though. Since it's variadic, we can just add it back in later without breaking any existing code. Also, the ordered quorum call implementation does not expose this option.

Issue #17, opened 2017-09-03
Make Manager and Configuration an interface to ease testing


s111:
This might or might not be a good idea, but I was thinking about having Manager and Configuration be interfaces. This should be easy enough to generate but we will end up having to change the implementing structs names, which should be fine. This might ease testing when your code is coupled to the network, i.e., Gorums, as you can mock out the Manager and Configurations.

I guess my main point with doing this was that we could mock out the network layer and easier create deterministic tests with, e.g., specific interleaving of messages, network delays and patriotions. This might end up being a rather large task. Just making the Manager and Configuration be interfaces might not be that big of a deal though.

meling:
Finally started to look into this. Sounds reasonable to me. I think it would be very useful if we can make the `Manager` and `Configuration` interfaces be part of the public API of Gorums, so that these interfaces need not be generated. This can also allow us to create helper functions, e.g. test helpers, for setting up Gorums.

meling:
Here is a proposed interface for `Configuration`.

```go
// Configuration represents a static set of nodes on which quorum remote
// procedure calls may be invoked.
type Configuration interface {
	Nodes() []*Node
	Size() int
	NewQuorumSpec() QuorumSpec
}
```

The `NewConfiguration` method should return the generated struct (not the interface), since then the interface does not need to contain the RPCs generated from the proto file.

The main issue is how can we access the `QuorumSpec`; we need generics.

meling:
When reviewing a PR a while back I noticed that we don't actually use the `Nodes()` method, nor the `NodeIDs()` method from within Gorums itself. The question is whether or not one or both of these methods will be useful as part of this interface.

PR #99, opened 2020-07-31
Pull lots of generated code out into the gorums package

### LINKED ISSUE #42: api: make wrapper(s) to simplify setup
Many use cases for Gorums currently repeat a lot of the same code, with small variations. I propose to make a helper function that is included in the generated `pb.go` file which can alleviate some of the boilerplate in some of the code and tests. Places to look for boilerplate:
- `config_qc_test.go:setup()`
- wherever we create `Manager`, `Configuration` and `QuorumSpec` objects
- especially, the handling of grpc credentials for secure connections.

(I'm still thinking if it is necessary, but I think we could do some simplifications for the developer, especially on the client side.)

### LINKED ISSUE #61: Allow creating multiple overlapping configurations
#### Problem
Currently, [`NewConfiguration`](https://github.com/relab/gorums/blob/master/dev/mgr.go#L222) only creates a new configuration, if no previous configuration with the same ids exists. Otherwise, it returns a reference to an already existing configuration. This raises the following issues:
1. It is not possible to create multiple configurations with identical nodes.
2. When recreating a configuration, the returned configuration will actually not use the QSpec provided.
3. It is not possible to garbage collect configurations.

I believe it is desirable to allow multiple configurations with the same set of nodes. This allows concurrent Quorum Calls to use different QSpecs. This allows maintaining call specific data in the QSpec without concurrency control.
Updating state in the QSpec can, for example, be useful to store, which replies have been successfully validated.

#### Fix
I want to propose to remove the `configs` map. Every call to `NewConfiguration` should indeed return a new configuration. The application can maintain them by reference. When the application does no longer references the configuration it should be garbage collected.
An interesting question is whether configuration ids should also be removed. Configuration ids allow to easily identify within and across processes if two configurations use the same underlying set of nodes.

What do you think @tormoder @meling

Raytar:
This PR attempts to move as much of the generated code out into the gorums package. What remains in the generated code is mostly wrappers that convert types between the application types and the generic `ProtoMessage` types.

Fixes #61 
Fixes #42 probably
Related #17


gorums.proto
```diff
@@ -12,6 +12,7 @@ extend google.protobuf.MethodOptions {
   bool multicast = 50003;
   bool correctable = 50004;
   bool unicast = 50005;
+  bool rpc = 50006;
```
meling, line 15: We can remove `ordered`. Also update numbering so it is consecutive. And maybe we should make the `per_node_arg` number `50100` to give more room for call type options.

node.go
```diff
@@ -0,0 +1,215 @@
...
+// Warning: This function may be removed in the future.
+var Port = func(n1, n2 *Node) bool {
+	p1, _ := strconv.Atoi(n1.Port())
+	p2, _ := strconv.Atoi(n2.Port())
+	return p1 < p2
+}
+
+// Error sorts nodes by their LastErr() status in increasing order. A
+// node with LastErr() != nil is larger than a node with LastErr() == nil.
+var Error = func(n1, n2 *Node) bool {
```
meling, line 210: Not ideal to expose this as `gorums.Error`. See other comment.

quorumcall.go
```diff
@@ -0,0 +1,70 @@
...
+	PerNodeArgFn   func(protoreflect.ProtoMessage, uint32) protoreflect.ProtoMessage
+	QuorumFunction func(protoreflect.ProtoMessage, map[uint32]protoreflect.ProtoMessage) (protoreflect.ProtoMessage, bool)
+}
+
+func QuorumCall(ctx context.Context, d QuorumCallData) (resp protoreflect.ProtoMessage, err error) {
+	msgID := d.Manager.nextMsgID()
+	// set up channel to collect replies to this call.
+	replyChan := make(chan *orderingResult, len(d.Nodes))
+	d.Manager.putChan(msgID, replyChan)
+	// and remove it when the call it scomplete
```
meling, line 24: Typo: `when the call completes`

quorumcall.go
```diff
@@ -0,0 +1,70 @@
+package gorums
+
+import (
+	"context"
+
+	"github.com/relab/gorums/ordering"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type QuorumCallData struct {
```
meling, line 10: Add doc comment to exported method and struct.

tests/Makefile
```diff
@@ -1,32 +1,29 @@
 # Tests that should be run each time
-RUNTESTS := ordering encoding metadata tls reconnect
+RUNTESTS := ordering metadata tls
```
meling, line 2: Should we re-add the `reconnect` test? Maybe it can be just tested directly in the root package, without any code gen.?

tests/encoding/encoding_test.go
```diff
@@ -1,334 +0,0 @@
-package encoding
```
meling, line 1: Don't we want these tests?

tests/metadata/metadata_test.go
```diff
@@ -62,7 +62,7 @@ func TestMetadata(t *testing.T) {
 		"id": fmt.Sprint(want),
 	})
 
-	mgr, err := NewManager(WithNodeList(addrs), WithMetadata(md), WithDialTimeout(1*time.Second), WithGrpcDialOptions(
+	mgr, err := NewManager(gorums.WithNodeList(addrs), gorums.WithMetadata(md), gorums.WithDialTimeout(1*time.Second), gorums.WithGrpcDialOptions(
```
meling, line 65: Line breaks.

tests/metadata/metadata_test.go
```diff
@@ -91,7 +91,7 @@ func TestPerNodeMetadata(t *testing.T) {
 		})
 	}
 
-	mgr, err := NewManager(WithNodeList(addrs), WithPerNodeMetadata(perNodeMD), WithDialTimeout(1*time.Second), WithGrpcDialOptions(
+	mgr, err := NewManager(gorums.WithNodeList(addrs), gorums.WithPerNodeMetadata(perNodeMD), gorums.WithDialTimeout(1*time.Second), gorums.WithGrpcDialOptions(
```
meling, line 94: Line breaks

tests/metadata/metadata_test.go
```diff
@@ -115,7 +115,7 @@ func TestCanGetPeerInfo(t *testing.T) {
 	addrs, teardown := gorums.TestSetup(t, 1, func() interface{} { return initServer(t) })
 	defer teardown()
 
-	mgr, err := NewManager(WithNodeList(addrs), WithDialTimeout(1*time.Second), WithGrpcDialOptions(
+	mgr, err := NewManager(gorums.WithNodeList(addrs), gorums.WithDialTimeout(1*time.Second), gorums.WithGrpcDialOptions(
```
meling, line 118: Line breaks

tests/ordering/order_test.go
```diff
@@ -80,11 +80,11 @@ func (q testQSpec) AsyncHandlerQF(_ *Request, replies map[uint32]*Response) (*Re
...
 	t.Helper()
 	addrs, closeServers := gorums.TestSetup(t, cfgSize, func() interface{} {
-		srv := NewGorumsServer()
-		srv.RegisterGorumsTestServer(&testSrv{})
+		srv := gorums.NewServer()
+		RegisterGorumsTestServer(srv, &testSrv{})
 		return srv
 	})
-	mgr, err := NewManager(WithNodeList(addrs), WithDialTimeout(100*time.Millisecond), WithGrpcDialOptions(
+	mgr, err := NewManager(gorums.WithNodeList(addrs), gorums.WithDialTimeout(100*time.Millisecond), gorums.WithGrpcDialOptions(
```
meling, line 87: Line breaks.

tests/reconnect/reconnect_test.go
```diff
@@ -1,60 +0,0 @@
-package reconnect
```
meling, line 1: Don't we want this?

tests/tls/tls_test.go
```diff
@@ -41,13 +41,13 @@ func TestTLS(t *testing.T) {
...
-		srv := NewGorumsServer(WithGRPCServerOptions(grpc.Creds(credentials.NewServerTLSFromCert(&tlsCert))))
-		srv.RegisterTLSServer(&testSrv{})
+		srv := gorums.NewServer(gorums.WithGRPCServerOptions(grpc.Creds(credentials.NewServerTLSFromCert(&tlsCert))))
+		RegisterTLSServer(srv, &testSrv{})
 		return srv
 	})
 	defer teardown()
 
-	mgr, err := NewManager(WithNodeList(addrs), WithDialTimeout(100*time.Millisecond), WithGrpcDialOptions(
+	mgr, err := NewManager(gorums.WithNodeList(addrs), gorums.WithDialTimeout(100*time.Millisecond), gorums.WithGrpcDialOptions(
```
meling, line 50: Line breaks

Raytar:
I've been thinking about the change done in <https://github.com/relab/gorums/pull/99/commits/8b3912e09d2d7f907996955d7642fa64d78e1709>, and I'm not sure if it's best to keep it or revert it.
To sum it up, the commit removes the context parameter from the unicast and multicast calltypes, and instead creates a timeout context internally before sending the message. This timeout can be set by using a manager option. I am not a fan of needing timeouts everywhere, which is why I am now reconsidering this commit.

These contexts are needed to abort the stream if it gets stuck (if the receiving node stops receiving messages or is very slow). The reason I removed the context parameter is that the code that calls the calltype does not know whether or not a message has been sent, making it difficult to cancel the context correctly.

Before this commit, the calling code would have to deal with the `cancel` function somehow.
One solution would be to ignore the `cancel`:

```go
ctx, _ := context.WithTimeout(context.Background(), 10 * time.Millisecond)
node.Unicast(ctx, msg)
```

But now we get a `govet` warning: 
```text
the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
```

Looking at the [source code](https://github.com/golang/go/blob/bc7e4d9257693413d57ad467814ab71f1585a155/src/context/context.go#L451) for the timeout context, the context is obviously canceled when the timeout happens, and so I assume that the code above does not actually leak a context after the timeout. However, it is not ideal to ignore the documentation.

Another solution would be to keep the `cancel` function and run it at a time when the message is assumed to be sent. For example, in HotStuff, votes are sent to the leader using a unicast. We could store the context cancel function until the next time the node votes, at which point it is safe to cancel the last context because it doesn't matter whether or not the previous vote was sent. This way we might be able to use a cancel context instead of a timeout context as well. It would look like this:

```go
func handlePropose(msg Proposal) {
  // prepare vote
  ...
  cancel() // cancel's the context from the previous proposal
  ctx, cancel = context.WithCancel(context.Background())
  leader.Vote(ctx, vote)
}
```

Another possible solution is to make the Unicast/Multicast methods in Gorums block until the message is actually sent. I haven't looked into how this could be done, but then the context can simply be canceled as soon as the method returns. But this might create a problem in HotStuff: Currently, a proposal is handled synchronously for each leader, and the votes are sent at the end of the handlers. But if the voting becomes a blocking function, then the node may not be able to handle another message from the proposer until the vote times out and unblocks.

I am leaning towards reverting this commit for now and implementing the second solution (canceling the previous context when the next message is being prepared) in HotStuff.

meling:
_This is a re-posting of @Raytar's comment that were unexpectedly deleted by a bug in GitHub._

I've been thinking about the change done in <https://github.com/relab/gorums/pull/99/commits/8b3912e09d2d7f907996955d7642fa64d78e1709>, and I'm not sure if it's best to keep it or revert it.

**Option 1:** To sum it up, the commit removes the context parameter from the unicast and multicast calltypes, and instead creates a timeout context internally before sending the message. This timeout can be set by using a manager option. I am not a fan of needing timeouts everywhere, which is why I am now reconsidering this commit.

These contexts are needed to abort the stream if it gets stuck (if the receiving node stops receiving messages or is very slow). The reason I removed the context parameter is that the code that calls the calltype does not know whether or not a message has been sent, making it difficult to cancel the context correctly.

Before this commit, the calling code would have to deal with the `cancel` function somehow.
One solution would be to ignore the `cancel`:

```go
ctx, _ := context.WithTimeout(context.Background(), 10 * time.Millisecond)
node.Unicast(ctx, msg)
```

But now we get a `govet` warning:
```text
the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
```

Looking at the [source code](https://github.com/golang/go/blob/bc7e4d9257693413d57ad467814ab71f1585a155/src/context/context.go#L451) for the timeout context, the context is obviously canceled when the timeout happens, and so I assume that the code above does not actually leak a context after the timeout. However, it is not ideal to ignore the documentation.

**Option 2:** Another solution would be to keep the `cancel` function and run it at a time when the message is assumed to be sent. For example, in HotStuff, votes are sent to the leader using a unicast. We could store the context cancel function until the next time the node votes, at which point it is safe to cancel the last context because it doesn't matter whether or not the previous vote was sent. This way we might be able to use a cancel context instead of a timeout context as well. It would look like this:

```go
func handlePropose(msg Proposal) {
  // prepare vote
  ...
  // cancels the context from the previous proposal.
  // ensures that the stream will be reconnected if it gets stuck,
  // so that the next vote can be put in the queue without blocking.
  cancel()
  ctx, cancel = context.WithCancel(context.Background())
  leader.Vote(ctx, vote)
}
```

**Option 3:** Another possible solution is to make the Unicast/Multicast methods in Gorums block until the message is actually sent. I haven't looked into how this could be done, but then the context can simply be canceled as soon as the method returns. But this might create a problem in HotStuff: Currently, a proposal is handled synchronously for each leader, and the votes are sent at the end of the handlers. But if the voting becomes a blocking function, then the node may not be able to handle another message from the proposer until the vote times out and unblocks.

I am leaning towards reverting this commit for now and implementing the second solution (canceling the previous context when the next message is being prepared) in HotStuff.

meling:
Option 2: I like this option for the HotStuff scenario, and similar scenarios, where the next call to the same method actually cancels/overrides any previous calls. Hence, this should work for scenarios like the `Propose` scenario in HotStuff and idempotent messages. The question I have is whether or not there are scenarios where that's not the desired behavior. 

Option 3: This might be possible now that we no longer depend on gRPC for this stuff; we might actually know when a message has been sent, which wasn't possible in gRPC. I haven't studied this in-depth.

PR #102, opened 2021-01-08
encoding: revert to using method names instead of numeric ids

Raytar:
When we have multiple gorums services in the same project, both services will register a codec with gRPC. Because both services use the same content subtype ("gorums"), one service will end up overwriting the codec of the other. Because the codecs contain the mapping of IDs to message types, one service may not be able to unmarshal any of its messages.

This PR fixes this problem by reverting back to using each method's full name (as defined by protobuf), instead of a numerical ID assigned by gorums at compile time (see #62). The unmarshaler now looks up the correct types based on the full name, and no information about the types is stored in the codec. Thus, multiple services can use the same codec for marshaling / unmarshaling.


meling:
Looks good to me! But I'm curious if it wouldn't be possible to use method IDs if they could be made unique somehow, e.g. maybe create separate method ID tables for each service, or similar?

Raytar:
Another option would be to use a unique content subtype for each service. That way each service can register its own codec (which can have its own list of method ids). But if we can't make that smaller than the method FullName, then there is no point, as the content subtype is also sent as part of the message.

PR #103, opened 2021-01-13
Make TestSetup use type safe ServerIface

meling:
This makes the TestSetup function take a type safe
function as argument instead of interface{}.


Raytar (approved):
LGTM

PR #104, opened 2021-01-13
Rename Future to Async

meling:
This PR renames the Future call type to Async to match the gorums.async option naming.


Raytar (approved):
LGTM

PR #108, opened 2021-01-16
Add receiveQueue.newCall()

meling:
This PR adds newCall to receiveQueue and adds benchmarks to show that it doesn't impact real performance in any detectable manner. 

A follow-up PR should remove the functions:
- nextMsgID
- putChan
- deleteChan

And the corresponding benchmarks that depend on these functions.


Raytar (approved):
LGTM

PR #113, opened 2021-02-04
Decoupling Node and Manager

meling:
This PR aims to decouple the Node type from Manager.
There are still a few places where the dependence is tight, but perhaps this is ok.

### DIFF
--- a/cmd/benchmark/main.go
+++ b/cmd/benchmark/main.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"runtime"
 	"runtime/pprof"
+	"runtime/trace"
 	"strings"
 	"syscall"
 	"text/tabwriter"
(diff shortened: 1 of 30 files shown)


cmd/benchmark/main.go
```diff
@@ -130,20 +128,6 @@ func main() {
 		defer pprof.StopCPUProfile()
 	}
 
-	if *traceFile != "" {
```
Raytar, line 133: Actually, this "tracing" code works fine. I was referring to the use of the `net/trace` package within Gorums. The `runtime/trace` package is different.

meling: Okay. I see. Do you mean that `grpc.EnableTracing` can be removed along with `trace.IsEnabled()` check line 195.

mgr_test.go
```diff
@@ -74,6 +79,39 @@ func TestManagerAddNode(t *testing.T) {
 		if err != nil && err.Error() != test.err {
 			t.Errorf("mgr.AddNode(%s, %d) = %s, expected %s", test.addr, test.id, err.Error(), test.err)
 		}
+	}
+}
 
+// Reusing tests/unresponsive/unresponsive.proto definitions
```
Raytar, line 85: Maybe we should have a dedicated package for this kind of testing?

cmd/benchmark/main.go
Raytar: Yes. But the `runtime/trace` stuff can stay

mgr_test.go
meling: Added new `tests/dummy/dummy.proto` to be used for such testing. Please let me know if you have suggestions for a better name.

Raytar (approved):

Issue #115, opened 2021-02-08
Make RPC and Unicast be methods on a single node configuration?


meling:
When thinking about certain aspects related to Node vs Configuration, I came to think about the following. Should we just drop support for invoking methods on a Node, and instead only allow invoking a method on a Configuration?? I'm thinking about our RPC and unicast call types that both work directly on a Node type. Here is what the code would look like:
```golang
// GRPCCall plain gRPC call; testing that Gorums can ignore these, but that
// they are added to the _grpc.pb.go generated file.
func (c *Configuration) GRPCCall(ctx context.Context, in *Request) (resp *Response, err error) {
	if c.Size() != 1 {
		return nil, fmt.Errorf("cannot invoke gRPC call on configuration with size %d", c.Size())
	}
	cd := gorums.CallData{
		Node:    c.Nodes()[0],
		Message: in,
		Method:  "dev.ZorumsService.GRPCCall",
	}
```
Instead of:
```golang
func (n *Node) GRPCCall(ctx context.Context, in *Request) (resp *Response, err error) {
	cd := gorums.CallData{
		Node:    n.Node,
		Message: in,
		Method:  "dev.ZorumsService.GRPCCall",
	}
```
Reason: We can avoid the following struct in the generated code, and things become easier to maintain (we need one less translation layer...):
```golang
type Node struct {
	*gorums.Node
}
```
@Raytar and @leandernikolaus  Thoughts?

Would could add a helper method to the `Configuration` type to return a single node `Configuration` object from an existing configuration.

Raytar:
I prefer to have separate types for `Node` and `Configuration`. It doesn't make sense to me that a "unicast" method would be implemented on a configuration of nodes, or vice versa.

Raytar:
Would it be possible to get rid of `Configuration` instead? Maybe we could use `type Configuration []*Node`?

meling:
Not done think about this, but in #116, there are now both `dev.Configuration` and `gorums.Configuration`, and `QSpec` is only defined on `dev.Configuration`.

I've also realized that we could define `func (n *Node) GRPCCall( ... )` in package `gorums` instead of making the `Node` object part of the `CallData` type. Similarly, other call types defined in `gorums` can now be defined as `func (c *Configuration) QuorumCall( ... )`, instead of making `Nodes []*gorums.Node` be part of the `QuorumCallData` type. Not sure if it is better or not, but we have the possibility to do it.

meling:
We will keep both Node and Configuration.

PR #116, opened 2021-02-09
Move Configuration to gorums package

meling:
- Sort nodes by ID also for WithNodeList option
- Migrated much of dev.Configuration to gorums.Configuration

### DIFF
--- a/config.go
+++ b/config.go
@@ -5,12 +5,9 @@ import (
 )
 
 // Configuration represents a static set of nodes on which quorum calls may be invoked.
-type Configuration struct {
-	nodes []*Node
-	errs  chan Error
-}
+type Configuration []*Node
 
-func NewConfiguration(mgr *Manager, ids []uint32) (*Configuration, error) {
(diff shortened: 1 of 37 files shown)


meling:
This PR builds on #114, so should be merged only after merging #114.

config.go
```diff
@@ -0,0 +1,79 @@
...
+	return ids
+}
+
+// Nodes returns the nodes in this configuration.
+func (c *Configuration) Nodes() []*Node {
+	return c.nodes
+}
+
+// Node returns a new configuration with a single node with the given identifier, if present.
+func (c *Configuration) Node(id uint32) (cfg *Configuration, found bool) {
```
Raytar, line 61: I assume this is related to #115 ?

meling: Yes. And as you point out in https://github.com/relab/gorums/issues/115#issuecomment-777571217 it may not be very intuitive. So probably we should remove it.

Raytar (approved):

PR #120, opened 2021-02-12
Pass nodes to Configuration instead of Manager

### LINKED ISSUE #30: Allow adding new nodes on the fly
Is there any reasons that the list of nodes need to be specified up front? As we are always creating a "new configuration", I don't see a problem with adding nodes at runtime to a new configuration.

For starters:
- Remove the `nodeAddrs` argument from `NewManager`.
- `Manager.NewConfiguration` creates connections to non-existent nodes on the fly, adding them to the manager for re-use.

meling:
This PR allows nodes to be passed to NewConfiguration instead of adding it to the manager first, and then adding it to the configuration. This makes it more intuitive and it also becomes easier to create new configurations from existing ones. We can now also compare two configurations via a new Equal method.

We provide tree ways to add nodes to NewConfiguration:
- WithNodeList: provide a list of string endpoints; IDs are assigned.
- WithNodeMap: provide a map of string endpoints to IDs; used for application defined IDs.
- WithNodeIDs: provide a list of node IDs for the new configuration; these IDs must already have been added via one of the other options.

Also worth noting is that NewManager no longer returns an error, avoiding error handling until NewConfiguration is called.

Fixes #30.


Raytar:
@meling can you regenerate `template_static.go` to update the comment from the last commit?

Raytar (approved):

meling:
> @meling can you regenerate `template_static.go` to update the comment from the last commit?

Done. It is ready to be reviewed, I guess. The caveat with this one is that it impacts the APIs, but I think now is the time to fix this, rather than later, when we have made a proper release.

PR #122, opened 2021-02-13
Prototyping WithQuorumSpec

meling:
This PR should not be merged. It was added for archival and prototyping purposes.


meling:
There are better designs.

Issue #112, opened 2021-02-02
Small user guide error


AndreaEsposit:
You are using an outdated version of the ``newManager`` function in the user guide: 

It should be 
``` mgr, err := pb.NewManager(pb.WithNodeList(addrs), ....) ```
 Instead of 
```mgr, err := pb.NewManager(addrs, ....) ```

meling:
Fixed by #123

PR #123, opened 2021-02-18
Revised user guide and other docs

### LINKED ISSUE #111: The protoc command given in the user guide does not include gorums.proto in the proto path
In the user guide we have this command:

```
protoc -I=. \
  --go_out=paths=source_relative:. \
  --gorums_out=paths=source_relative:. \
  storage.proto
```

It does not work because the `-I` parameter does not include the path to the gorums.proto file.

One possible solution is to change it to

```
protoc -I=$(go list -m -f {{.Dir}} github.com/relab/gorums):. \
  --go_out=paths=source_relative:. \
  --gorums_out=paths=source_relative:. \
  storage.proto
```

This should work as long as the user has set up a go module and has added gorums as a dependency.

meling:
This PR updates the user guide and other documentation to follow the most recent API changes and removal of the gorums.ordered option.

Fixes #111, #112.


doc/user-guide.md
```diff
@@ -160,15 +158,11 @@ func (srv *storageSrv) Write(_ context.Context, req *State, ret func(*WriteRespo
 
 There are some important things to note about implementing the server interfaces:
 
-* Reply messages should be returned using the `ret` function.
-* The handlers are run synchronously, in the order that messages are received.
-  This means that a long-running handler will prevent other messages from being handled.
+* Reply messages must be returned using the `ret` function.
+* The handlers run synchronously, in the order messages are received.
```
Raytar, line 162: Should probably add that the ordering only applies to messages between a single node and the server. The order of execution between different nodes is not guaranteed. For example:

> The handlers run in the order that messages are received per node. So all messages from a single node are executed in the order they are received, but messages from two different nodes may not be. Handlers run synchronously, and hence a long-running handler will prevent other messages from being handled.

doc/user-guide.md
```diff
@@ -181,12 +175,16 @@ There are some important things to note about implementing the server interfaces
...
   This may change with other call types in the future.
 
-For more information about message ordering and why we use channels instead of `return` in our handlers, read [ordering.md](./ordering.md)
+For more information about message ordering and why we use channels instead of `return` in our handlers, read [ordering.md](./ordering.md).
 
-To start the server, we need to create a *listener* and a *GorumsServer*, and then register our server implementation with the GorumsServer:
+To start the server, we need to create a *listener* and a *GorumsServer*, and then register our server implementation:
 
 ```go
 func ExampleStorageServer(port int) {
```
Raytar, line 190: This code needs to be updated:  
 

```go
func ExampleStorageServer(port int) {
  lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
  if err != nil {
    log.Fatal(err)
  }
  gorumsSrv := gorums.NewServer()
  srv := storageSrv{state: &State{}}
  RegisterStorageServer(gorumsSrv, &srv)
  gorumsSrv.Serve(lis)
}
```

doc/user-guide.md
```diff
@@ -223,40 +222,39 @@ import (
 )
 
 func ExampleStorageClient() {
+  mgr := NewManager(
+    WithDialTimeout(500*time.Millisecond),
```
Raytar, line 226: Should be `gorums.WithDialTimeout` and `gorums.WithGrpcDialOptions`

doc/user-guide.md
```diff
@@ -268,7 +266,7 @@ We can now invoke the Write RPC on each `node` in the configuration:
   }
 
   // Invoke Write RPC on all nodes in config
-  for _, node := range allNodesConfig.Nodes() {
+  for _, node := range allNodesConfig {
```
Raytar, line 269: Is this actually possible now? I thought only `gorums.Configuration` was a slice, not the generated `dev.Configuration`.

doc/user-guide.md
```diff
@@ -268,7 +266,7 @@ We can now invoke the Write RPC on each `node` in the configuration:
...
     respons, err := node.Write(context.Background(), state)
```
Raytar, line 270: Typo: respons

Although I think we should use `reply` instead since we use `reply` in other parts of the document.

doc/user-guide.md
```diff
@@ -278,33 +276,31 @@ We can now invoke the Write RPC on each `node` in the configuration:
...
-Instead of invoking an RPC explicitly on all nodes in a configuration, Gorums allows users to invoke the RPC as a quorum call on the configuration.
-If an RPC is invoked as a quorum call, Gorums will invoke the RPCs on all nodes in the configuration in parallel, and collect and process the replies.
+Instead of invoking an RPC explicitly on all nodes in a configuration, Gorums allows users to invoke a _quorum call_ via a method on the `Configuration` type.
+If an RPC is invoked as a quorum call, Gorums will invoke the RPCs on all nodes in parallel and collect and process the replies.
 
-For the Gorums plugin to generate quorum calls we have to specify the `quorumcall` option for our RPC methods in the proto file, as shown below:
+For the Gorums plugin to generate quorum calls we need to specify the `quorumcall` option for our RPC methods in the proto file, as shown below:
 
 ```protobuf
 service QCStorage {
```
Raytar, line 289: This requires `gorums.proto` to be imported, and that has not been mentioned at this point.

doc/user-guide.md
```diff
@@ -401,19 +399,17 @@ func ExampleStorageClient() {
     "127.0.0.1:8082",
   }
 
-  mgr, err := NewManager(WithNodeList(addrs), WithDialTimeout(50*time.Millisecond), WithGrpcDialOptions(
-    grpc.WithBlock(),
-    grpc.WithInsecure(),
+  mgr := NewManager(
+    WithDialTimeout(50*time.Millisecond),
```
Raytar, line 403: should be `gorums.WithDialTimeout`, `gorums.WithGrpcDialOptions` and `gorums.WithNodeList`

doc/user-guide.md
```diff
@@ -1,73 +1,72 @@
...
 
-First, we need version 3 of `protoc`, the Protocol Buffers Compiler.
-Installation of this tool is OS/distribution specific.
-See <https://github.com/google/protobuf/releases> and <https://developers.google.com/protocol-buffers> for details.
+1. Install version 3 of `protoc`, the Protocol Buffers Compiler.
+   Installation of this tool is OS/distribution specific.
 
-Second, we need to install the [Go code generator](https://github.com/protocolbuffers/protobuf-go) for `protoc`.
-It can be installed with the following command:
+   On macOS with Homebrew you can use:
```
Raytar, line 18: Note: This works on Linux too

doc/user-guide.md
```diff
@@ -124,12 +122,12 @@ type Storage interface {
 }
 ```
 
-**Note:** For a real use case, you may decide to keep the `.proto` file and the generated `.pb.go` files in a separate directory/package, and import that package (the generated Gorums API) into to your main application.
+**Note:** For a real use case, you may decide to keep the `.proto` file and the generated `.pb.go` files in a separate directory/package and import that package (the generated Gorums API) into your application.
```
Raytar, line 125: Maybe it would be better if this guide used that approach instead?

Raytar (changes requested):

doc/user-guide.md
meling: Using homebrew? So I can just say On Linux/macOS with Homebrew?

doc/user-guide.md
meling: Opened issue #124 to address this in the future.

doc/user-guide.md
Raytar: Yes. Homebrew can be used on Linux. It may be preferable because the `protoc` version included in a distro's package manager might be outdated, and it is easier than installing it from a tarball.

PR #126, opened 2021-03-04
Type safe WithQuorumSpec

### LINKED ISSUE #121: Pass QuorumSpec via WithQuorumSpec
I have an idea to get rid of the `QuorumSpec` as a separate argument to:
```go
func (m *Manager) NewConfiguration(qspec QuorumSpec, opts ...gorums.ConfigOption) (c *Configuration, err error) {
```
Instead we can do:
```go
func (m *Manager) NewConfiguration(opts ...gorums.ConfigOption) (c *Configuration, err error) {
```
And pass the `qspec` instance via `WithQuorumSpec(qspec)`.

This `WithQuorumSpec()` must accept an `interface{}`, but we can just do a `qspec.(QuorumSpec)` before storing it in the `dev.Configuration` struct. I like this approach because not all configurations need a `QuorumSpec`, and it makes the API a bit cleaner.

There is only a small snag, in that the `gorums.NewConfiguration` has no concept of `QuorumSpec`, and it would be a mistake to add it since `QuorumSpec` is a generated interface specific to quorum calls defined in the application's proto file.

I've considered two approaches to work around this snag.
1. Let `gorums.NewConfiguration` receive and process `WithQuorumSpec` and just return it as an `interface{}`, as a separate return value.
2. Make a new function `gorums.GetQSpec()` that also processes `ConfigOption` types, but only returns the `qspec`. See code below.

I don't really like alternative 1 above because it makes things more coupled than they should.

What do you think about alternative 2? Are there any other alternatives?

```go
func GetQSpec(opts ...ConfigOption) interface{} {
	o := newConfigOptions()
	for _, opt := range opts {
		opt(&o)
	}
	return o.qspec
}
```
Using it from `dev.NewConfiguration` would look something like this:
```go
	if qspec, ok := gorums.GetQSpec(opts...).(QuorumSpec); !ok {
		return nil, errors.New("QuorumSpec error")
	} else {
		c.qspec = qspec
	}
	return c, nil
```

PS: This is probably referring to code as it looks in PR #120.

meling:
This is a new attempt at making the passing of the `QuorumSpec` argument more flexible. `NewConfiguration` now accepts 1 or 2 arguments only. One must be a `gorums.NodeListOption`, and the other *may* be a `QuorumSpec` if provided.

My initial idea was that the `QuorumSpec` should be passed via the `WithQuorumSpec` function. However, as it turns out, since I added the `gorums.ConfigOption` empty interface to the `QuorumSpec` interface, it seems that we no longer need this function and can pass the `QuorumSpec` directly. This means the API is compatible with existing code.

Note that I added forwarding methods for the `WithNodeList`, `WithNodeMap`, and `WithNodeIDs` in the generated code. However, it is not clear if these are really useful. We could instead just use the `gorums.With*` implementations. Especially if the generated versions of the `With` functions are placed in a separate package, e.g., `proto`, from the code where we call `NewConfiguration`. Then we would have to call them as `proto.With`, which isn't much better than `gorums.With`.

Fixes #121.


meling:
PS: I removed again the `dev.With` functions mentioned in the first comment.

Raytar:
Looks good! The only problem is that failing to provide a `QuorumSpec` when required results in a panic. I have been trying to figure out how to check this, and I came up with this creative use of interfaces and type assertions:

```go
// return an error if the QuorumSpec interface is not empty and no implementation was provided.
var test interface{} = struct{}{}
if _, empty := test.(QuorumSpec); !empty && c.qspec == nil {
	return nil, fmt.Errorf("missing required QuorumSpec")
}
```

Raytar (approved):

PR #129, opened 2021-03-04
Improved version handling

### LINKED ISSUE #128: Prevent codegen errors when protoc-gen-gorums version differs from imported gorums version
If the version of `protoc-gen-gorums` differs from the gorums version imported by a module, we may encounter problems with code generation. See: https://github.com/relab/hotstuff/issues/28

Can we detect this in the code generator and output warnings/errors?

meling:
Adds version comment in generated files and updates `version.go` to match tagged version.

Follow-up commits should add checks that generated code is used together with the appropriate gorums library version.

Furthermore, we should figure out a way to update `internal/version.go` based on the git tag or something similar.
See https://github.com/golang/go/issues/29228

Fixes #128.


meling:
This adopts the same version checking as protoc-gen-go. Not sure if we could make this easier with git tags, but this is at least a first stab.

Raytar (approved):
Looks good!

Issue #109, opened 2021-01-16
Move call type implementations to new package calltype


meling:
Moving the call types to a new package `calltype` can help with code maintainability. @Raytar Thoughts?

Raytar:
Yes, that sounds like a good idea. It would also help with making the API clearer since the call types are meant to be used by generated code.

PR #131, opened 2021-03-18
Add support for multiple streams (channels) per Node

Raytar:
This PR attempts to make it possible to have multiple `Channels` (formerly `orderedNodeStream`) per `Node`. Currently, the channel for a call can be changed from the default by passing either a `*gorums.Channel` or a `[]*gorums.Channel` as a `ConfigOption`.


meling:
@Raytar What do you want to do with this PR? There are some nice things to keep in this, even if we don't need the channels. It would be nice to bring these things into master before giving another go at #109.

PR #132, opened 2021-03-18
Add/remove convenience methods on Configuration

### LINKED ISSUE #28: Add convenience methods to the configuration object
What do you think of this change to the configuration object?

- Add `add/remove/replace` convenience methods to `Configuration`, i.e., c.add(x, q) is identical to `Manager.NewConfiguration(append(c.NodeIDs(), x), q)` and similar for `remove` and `replace`. (I don't know if `replace` is actually useful, I was thinking if we wanna change the IP of a node on the fly, but that can be achieved by just using hostnames and DNS).
  - (We could maybe drop the need for a new quorum spec when using the `add/remove` methods by adding a `OnChange` method to the `QuorumSpec`, i.e., this method would be implemented and would be responsible for changing the read and write quorum size and etc.., when a new configuration is created from an existing configuration with that QuorumSpec)

meling:
This adds methods Add, Remove, AddNodes, RemoveNodes to gorums.Configuration and correspondingly to dev.Configuration via struct embedding and method forwarding as needed.

This doesn't actually implement what was proposed in #28 exactly but gives the flexibility to add/remove nodes/configurations to/from an existing configuration. We still need to use the mgr.NewConfiguration method to actually produce the new configuration, so that we can pass in the QuorumSpec.

Closes #28.


meling:
The current design in this PR allows this:
```golang
	c3, err := mgr.NewConfiguration(
		newQSpec(c1.Size()+c2.Size()),
		c1.Add(c2),
	)
```

However, we could do this instead:
```golang
	c3, err := c1.Add(c2, newQSpec(c1.Size()+c2.Size()))
```

This is clearly shorter, and it doesn't require access to the `mgr` to create new configurations from existing ones.

The benefit of the first approach is that we could also do:
```golang
	newNodeList := c1.Add(c2)
	c3, err := mgr.NewConfiguration(
		newQSpec(newNodeList.Size()),
		newNodeList,
	)
```

It is also more clear that `mgr.NewConfiguration` returns a `Configuration`, as opposed to `c1.Add(c2...)`.

As these are really just convenience methods, I prefer to keep only one `Add` method. I'm leaning towards the current approach, but want to hear other opinions. @Raytar @leandernikolaus @s111 @tormoder

meling:
Alternative naming for `Add`:
```golang
proto.Add(c1, c2)
proto.With(c1, c2)
proto.WithConfigurations(c1, c2)
proto.Combine(c1, c2)
proto.Merge(c1, c2)

c1.Add(c2)
c1.With(c2)
c1.WithConfiguration(c2)
c1.Combine(c2)
c1.Merge(c2)
```

tests/config/config_test.go
```diff
@@ -0,0 +1,117 @@
...
+		t.Fatal(err)
+	}
+	teardown = func() {
+		mgr.Close()
+		closeServers()
+	}
+	return cfg, teardown
+}
+
+func TestConfig(t *testing.T) {
```
Raytar, line 70: Some comments explaining the test would be helpful here I think. I looks like the `setup()` function is creating a new configuration with new nodes, such that each configuration created with `setup()` contain separate sets of nodes? Then you are testing the `With()` and `Without()` on these configurations?

Raytar (changes requested):

Raytar:
What about `And()` and `Except()`?

```go
// c3 will contain the nodes of c1 and c2
c3, err := mgr.NewConfiguration(c1.And(c2))
// c3 will contain the nodes of c3 except the nodes of c2
c4, err := mgr.NewConfiguration(c3.Except(c2))
```

Raytar (approved):
LGTM!

PR #144, opened 2021-06-05
node: add a function to get the hostname / IP address only

Raytar:
This adds a method to Nodes that returns the hostname or IP address of the node, without the port number.

### DIFF
--- a/node.go
+++ b/node.go
@@ -110,6 +110,15 @@ func (n *Node) Address() string {
 	return nilAngleString
 }
 
+// Host returns the network host of n.
+func (n *Node) Host() string {
+	if n == nil {
+		return nilAngleString
+	}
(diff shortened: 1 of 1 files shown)


meling:
How about replacing the existing `Port` method with a `func (n *Node) HostPort() (host, port string)`. The port could also be `int` or `uint16`; the stdlib is somewhat schizophrenic about how to represent ports, but most uses are `string`, and then some `int` and I think only one use of `uint16`, even though that's the correct representation.

But on the other hand, it can be a bit awkward to use such a method, if you are only interested in one of the return values, so I'm also okay with the suggested `Host()` method.

Raytar:
I don't have a use case for the `HostPort()` function, so I'm thinking its best to keep the old `Port()` function. When I need both the host and port together, I usually need them to be in the same string anyway.

PR #148, opened 2021-06-28
Fix Correctable streams

Raytar:
This PR fixes correctable streams and adds tests. Support for streams is implemented by adding a "streaming" parameter to the client side `channel.enqueue()` function, such that it won't automatically delete the "response router" after the first message is received. Instead, this PR adds a `channel.deleteRouter` function that is called once the correctable has completed.


meling (approved):
LGTM!

PR #149, opened 2021-12-17
Fix go report card issues

### LINKED ISSUE #139: Fix lint issues found by Go Report Card and add badge
https://goreportcard.com/report/github.com/relab/gorums

Raytar:
Fixes #139


correctable.go
```diff
@@ -73,6 +79,7 @@ func (c *Correctable) set(reply protoreflect.ProtoMessage, level int, err error,
 	}
 }
 
+// CorrectableCallData contains data for making a correctable quorum call.
```
meling, line 82: Should we add a comment like this to these types of structs:
```text
This struct should only be used by generated code.
```

internal/version/version.go
```diff
@@ -5,6 +5,7 @@ import (
 	"strings"
 )
 
+// Semantic version number parts.
```
meling, line 8: Perhaps we should check if the protobuf people have revised their corresponding code; that is, there is a TODO item and some links below in the `String()` function that maybe have been fixed? Didn't check.

cmd/benchmark/profiling.go
```diff
@@ -0,0 +1,103 @@
...
+func StartTrace(tracePath string) (stop func() error, err error) {
+	traceFile, err := os.Create(tracePath)
+	if err != nil {
+		return nil, err
+	}
+	if err := trace.Start(traceFile); err != nil {
+		return nil, err
+	}
+	return func() error {
+		// stop trace
```
meling, line 56: Remove this comment (next line says it all ;-)

cmd/benchmark/profiling.go
```diff
@@ -0,0 +1,103 @@
...
+func StartCPUProfile(cpuProfilePath string) (stop func() error, err error) {
+	cpuProfile, err := os.Create(cpuProfilePath)
+	if err != nil {
+		return nil, err
+	}
+	if err := pprof.StartCPUProfile(cpuProfile); err != nil {
+		return nil, err
+	}
+	return func() error {
+		// stop cpu profile
```
meling, line 21: Remove this comment (next line says it all ;-)

cmd/benchmark/profiling.go
```diff
@@ -0,0 +1,103 @@
...
+	}, nil
+}
+
+// StartProfilers starts various profilers and returns a function to stop them.
+func StartProfilers(cpuProfilePath, memProfilePath, tracePath string) (stopProfile func() error, err error) {
+	nilFunc := func() error { return nil }
+
+	var (
+		cpuProfileStop = nilFunc
+		traceStop      = nilFunc
```
meling, line 72: Seems to me that `nilFunc` could be replaced with:
```go
	var (
		cpuProfileStop   func() error
		traceStop  func() error
	)
```
Or am I missing something?

encoding.go
```diff
@@ -19,6 +20,7 @@ const (
 	responseType
 )
 
+// Message encapsulates a request message and metadata.
```
meling, line 23: Are these structs also meant to be used by generated code?

meling (approved):
Looks good; just a few minor things. Hence I approve directly. Feel free to fix the remaining issues and merge.

cmd/benchmark/profiling.go
Raytar: Actually, I want the function pointers to be non-nil, and thus I used `nilFunc` as a placeholder function.

PR #153, opened 2021-12-23
fix benchmark

Raytar:
My last PR (#149) reordered parts of the benchmark main function. I guess I forgot to test that it still works (it doesn't). Anyway, here's the fix.


meling (approved):
LGTM!

PR #155, opened 2021-12-31
Custom subconfigurations for a tree

meling:
- Added Configuration.Contains(nodeID)
- Changed WithNodeIDs(ids []uint32) to ids ...uint32
- Custom subconfigurations for a tree


Raytar:
Looking at the `SubConfigurations` function, the `mgr` parameter could easily be replaced by an interface returning the raw manager instead of passing the raw manager directly:

```go
// AsRawManager provides access to the underlying RawManager of a generated Manager.
type AsRawManager interface {
	// AsRaw returns a pointer to the underlying RawManager.
	AsRaw() *RawManager
}
```

```diff
 // Derive subconfigurations from the manager's base configuration.
-func SubConfigurations(mgr *RawManager, opt SubConfigOption) ([]RawConfiguration, error) {
+func SubConfigurations(mgr AsRawManager, opt SubConfigOption) ([]RawConfiguration, error) {
        if opt == nil {
                return nil, ConfigCreationError(fmt.Errorf("missing required subconfiguration option"))
        }
-       return opt.subConfig(mgr)
+       return opt.subConfig(mgr.AsRaw())
 }
```

This makes it possible to pass a generated manager as the argument to the function, but then there is the problem of converting the raw configurations returned by the function back into generated configurations. Do you have any ideas on how to make this work?

Raytar:
Of course, one solution is to move this function into the generated code, but that would necessitate changes to the `SubConfigOption` interface and further increase the number of APIs exposed by gorums that are not intended to be used outside of generated code.

meling:
Superseded by #328

PR #159, opened 2022-02-27
Configuration creation with configurable replica count and wait time

hanish520:
Tests to form configuration with a configurable number of replicas and wait time.


tests/all2all/alltoallconfig_test.go
```diff
@@ -0,0 +1,118 @@
...
+			if err := replica.serve(); err != nil {
+				errChan <- fmt.Errorf("failed to serve at %q: %w", replica.address, err)
+			}
+		}()
+	}
+	for startedCnt < numReplicas {
+		<-startedChan
+		startedCnt++
+	}
+	close(startedChan)
```
Raytar, line 69: Closing this channel here is wrong and unnecessary. Channels should be closed by senders only (see https://pkg.go.dev/builtin#close), and closing a channel is only necessary if the receiver needs to know that all data has been sent.

tests/all2all/alltoallconfig_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+// createReplicas returns a slice of replicas.
+// The function waits for all serve goroutines to start and one additional
+// second to allow the servers to start before returning.
+func createReplicas(numReplicas int) ([]*replica, error) {
+	replicas := make([]*replica, 0)
+	errChan := make(chan error, numReplicas)
+	startedChan := make(chan struct{}, numReplicas)
+	startedCnt := 0
+	defer close(errChan)
```
Raytar, line 44: This channel does not need to be closed either.

tests/all2all/alltoallconfig_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for i := 1; i <= numReplicas; i++ {
```
Raytar, line 45: If we initialize replicas with a size, i.e. `replicas := make([]*replica, numReplicas)`, this for loop can be simplified to

```go
for i in range replicas {
  ...
  replicas[i] = &replica
  ...
}
```

tests/all2all/alltoallconfig_test.go
```diff
@@ -0,0 +1,118 @@
...
+		RegisterSampleServer(replica.server, replica)
+		replicas = append(replicas, &replica)
+		go func() {
+			startedChan <- struct{}{}
+			if err := replica.serve(); err != nil {
+				errChan <- fmt.Errorf("failed to serve at %q: %w", replica.address, err)
+			}
+		}()
+	}
+	for startedCnt < numReplicas {
```
Raytar, line 65: This for loop can be simplified to

```go
for range replicas {
  <-startedChan
}
```

tests/all2all/alltoallconfig_test.go
```diff
@@ -0,0 +1,118 @@
...
+		for _, replica := range replicas {
+			replica.stopServer()
+		}
+	}()
+	nodeMap := make(map[string]uint32)
+	for _, replica := range replicas {
+		nodeMap[replica.address] = replica.id
+	}
+	for _, replica := range replicas {
+		replica.createConfiguration(nodeMap)
```
Raytar, line 32: The error returned by this method is not checked.

meling: Interestingly, this seems to bring back the bug we are fighting with here. If we run the code with:
```sh
$ go test -v -run TestAllToAllConfiguration -replicas 10
=== RUN   TestAllToAllConfiguration
--- PASS: TestAllToAllConfiguration (1.06s)
```
vs 
```sh
go test -v -run TestAllToAllConfiguration -replicas 11
=== RUN   TestAllToAllConfiguration
    alltoallconfig_test.go:33: could not create configuration: connection failed for addr: 127.0.0.1:65204: dialing node failed: context deadline exceeded
--- FAIL: TestAllToAllConfiguration (1.16s)
```
And this is consistently working for up to 10 replicas, but not for 11 or above.

meling: I added `TestGrpcDial` to try to understand better what's going on. This version fails consistently with 84 replicas (but not with fewer on my machine that is):
```sh
go test -v -run TestGrpc -replicas 84
=== RUN   TestGrpcDial
    alltoallconfig_test.go:133: context deadline exceeded
--- FAIL: TestGrpcDial (2.05s)
```

Raytar: I can't reproduce this on my machine

```
❯ ./all2all.test -test.v -test.run="TestGrpcDial" -replicas=10000
=== RUN   TestGrpcDial
--- PASS: TestGrpcDial (4.84s)
PASS
```

It only fails for me when the kernel runs out of ports 😅 

```
❯ ./all2all.test -test.v -test.run="TestGrpcDial" -replicas=50000 
=== RUN   TestGrpcDial
    alltoallconfig_test.go:123: listen tcp 127.0.0.1:0: bind: address already in use
--- FAIL: TestGrpcDial (9.77s)
FAIL
```

PR #160, opened 2022-03-03
server: add a ConnectCallback option

Raytar:
This option makes it possible to add a callback function that is called whenever a node connects or reconnects to the server. This thus makes it possible to inspect the node's metadata before it sends any RPCs.

### DIFF
--- a/server.go
+++ b/server.go
@@ -8,7 +8,6 @@ import (
 	"github.com/relab/gorums/ordering"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


meling:
This looks interesting and useful. But do you have a specific use case for this?

Raytar:
Yes! In hotstuff I am planning to store BLS proof-of-possession signatures in the gRPC metadata, and then use this callback to obtain the proof once a replica connects.

Raytar:
The metadata is obtained from the `NodeStream` stream context and passed into the callback function. The same context can also be used to obtain peer information. Maybe we should just pass the context to the callback function instead, and let that function decide what information to extract from the context?

meling (changes requested):
Looks good to me!! Maybe you want fix one minor issue I added. Other than that we can merge this.

PR #167, opened 2024-02-01
fix: node RPCCall returns nil instead of error

vidarandrebo:
Fixed a bug where an error was not returned from node.RPCCall
Added tests for all return paths in RPCCall function.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -18,7 +18,7 @@ type CallData struct {
 // RPCCall executes a remote procedure call on the node.
 //
 // This method should be used by generated code only.
-func (n *RawNode) RPCCall(ctx context.Context, d CallData) (resp protoreflect.ProtoMessage, err error) {
+func (n *RawNode) RPCCall(ctx context.Context, d CallData) (protoreflect.ProtoMessage, error) {
 	md := &ordering.Metadata{MessageID: n.mgr.getMsgID(), Method: d.Method}
 	replyChan := make(chan response, 1)
 	n.channel.enqueue(request{ctx: ctx, msg: &Message{Metadata: md, Message: d.Message}}, replyChan, false)
@@ -26,10 +26,10 @@ func (n *RawNode) RPCCall(ctx context.Context, d CallData) (resp protoreflect.Pr
 	select {
 	case r := <-replyChan:
 		if r.err != nil {
-			return nil, err
+			return nil, r.err
 		}
 		return r.msg, nil
 	case <-ctx.Done():
-		return resp, ctx.Err()
+		return nil, ctx.Err()
 	}
 }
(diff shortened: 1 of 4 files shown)


rpc.go
```diff
@@ -18,18 +18,18 @@ type CallData struct {
...
-func (n *RawNode) RPCCall(ctx context.Context, d CallData) (resp protoreflect.ProtoMessage, err error) {
+func (n *RawNode) RPCCall(ctx context.Context, d CallData) (protoreflect.ProtoMessage, error) {
 	md := &ordering.Metadata{MessageID: n.mgr.getMsgID(), Method: d.Method}
 	replyChan := make(chan response, 1)
 	n.channel.enqueue(request{ctx: ctx, msg: &Message{Metadata: md, Message: d.Message}}, replyChan, false)
 
 	select {
 	case r := <-replyChan:
 		if r.err != nil {
-			return nil, err
```
vidarandrebo, line 29: err is always nil here

rpc.go
```diff
@@ -18,18 +18,18 @@ type CallData struct {
...
+			return nil, r.err
 		}
 		return r.msg, nil
 	case <-ctx.Done():
-		return resp, ctx.Err()
```
vidarandrebo, line 33: resp is always nil here

rpc.go
```diff
@@ -18,18 +18,18 @@ type CallData struct {
...
 // This method should be used by generated code only.
-func (n *RawNode) RPCCall(ctx context.Context, d CallData) (resp protoreflect.ProtoMessage, err error) {
+func (n *RawNode) RPCCall(ctx context.Context, d CallData) (protoreflect.ProtoMessage, error) {
 	md := &ordering.Metadata{MessageID: n.mgr.getMsgID(), Method: d.Method}
 	replyChan := make(chan response, 1)
 	n.channel.enqueue(request{ctx: ctx, msg: &Message{Metadata: md, Message: d.Message}}, replyChan, false)
 
 	select {
 	case r := <-replyChan:
 		if r.err != nil {
```
meling, line 28: Here we can simply return without the proceeding `if`:
```go
return r.msg, r.err
```
Or did I miss something? The tests still pass with this version.

rpc_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+}
+
+func initServer() *gorums.Server {
+	srv := gorums.NewServer()
+	dummy.RegisterDummyServer(srv, &testSrv{})
+	return srv
+}
+
+func rpcTestMgr() *dummy.Manager {
```
meling, line 106: Maybe a better name for this helper would be `gorumsTestMgr` or something?

meling (changes requested):
Looks great! A few minor things.

meling (approved):

PR #175, opened 2024-02-13
Latency recording per node experiment

meling:
This needs to have some tests; maybe using localhost with some extra delays imposed.


meling:
Superseded by #245

PR #179, opened 2024-03-13
feat: Create config regardless of connection failures

aleksander-vedvik:
A config can be created regardless of a connection to the nodes have failed or not. Instead the channel will try to connect (only once) to the node for each message.

The channel needs to know whether a connection has been established. If not, it will dial the node. Otherwise, it will try to create a stream to the node.


channel.go
```diff
@@ -50,29 +51,39 @@ type channel struct {
...
 		rand:            rand.New(rand.NewSource(time.Now().UnixNano())),
 		responseRouters: make(map[uint64]responseRouter),
+		connEstablished: false,
 	}
 }
 
 func (c *channel) connect(ctx context.Context, conn *grpc.ClientConn) error {
-	var err error
 	c.parentCtx = ctx
+	go c.sendMsgs()
```
aleksander-vedvik, line 71: It is important to start the goroutine regardless of a successful connection or not. This is because we need to have a listener on the c.sendQ channel. Otherwise, c.enqueue() (used in e.g. QuorumCall and Multicast) will block indefinitely. The reconnection logic is also implemented in c.sendMsgs(). The goroutine will be stopped when the node is closed, preventing leaking goroutines.

node.go
```diff
@@ -61,28 +61,48 @@ func NewRawNodeWithID(addr string, id uint32) (*RawNode, error) {
...
+	return nil
+}
+
+// creates a context that governs the channel. It is
+// used to stop all channel goroutines and the NodeStream.
+//
+// this method should be run for each connection to ensure
+// fresh contexts. Reusing contexts could result in reusing
+// a cancelled context.
+func (n *RawNode) ctxSetup() context.Context {
```
meling, line 97: What about naming it: `newContext`?

Doc comment:
```go
// newContext returns a new context for this node's channel.
// This context is used by the channel implementation to stop
// all goroutines and the NodeStream, when the context is canceled.
//
// This method must be called for each connection to ensure... (rest as is)
```
(or something like this??)

meling (changes requested):
Thanks for this nice work! I've reviewed this without any experimentation; our current tests are passing. I would feel much more confident about approving this if you could add some tests that will fail on the current master branch, but passes with this PR.

Some of my comments are nits that aren't important for the semantics, but will help to improve code quality and style.

aleksander-vedvik:
Have added a few tests and more documentation. There are now two ways in which the channel will try to reconnect to a node:
1. If a connection has never been established: It will try to reconnect when sending a message. (see c.sendMsgs)
2. If a connection has been successfully established: It will in addition to point 1 try to reconnect in the background by using a back off strategy. (see c.recvMsgs)

Had to add adjust when locking happened in c.reconnect() since sendMsgs and recvMsgs are running in two different goroutines. Now, it will only lock when creating the NodeStream.

Regarding the tests:
The last test is supposed to represent a server first being offline, then online, and then offline again. To do this, it is necessary to use unexported methods on the channel, meaning it is not possible to use "[black box testing](https://pkg.go.dev/testing)". As a result, I created some custom proto types in tests/mock which is not dependent on gorums (to prevent cyclic imports). However, I suspect that gorums uses a custom encoding for the proto messages (placed in the init() function in each generated gorums proto file), causing an error when unmarshalling a non-gorums proto message. I might be wrong in this assumption, but for now none of the test are actually checking the response at the server-side because it fails to unmarshal the message.

channel_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if len(mgr.Nodes()) < 1 {
+		t.Fatal("the node was not added to the configuration")
+	}
+	if node.conn == nil {
+		t.Fatal("connection should not be nil when NOT using WithBlock()")
+	}
+}
+
+func TestChannelReconnection(t *testing.T) {
+	if encoding.GetCodec(ContentSubtype) == nil {
```
meling, line 114: I tried without this, and it didn't seem to impact things (tests passed), but I'm unsure why. It will be included in `_gorums.pb.go` files, so if you use `NewManager` from such a generated file, it will be initialized with the `init()` function. Maybe you could tweak and reuse the `tests/dummy/dummy.proto` file instead of adding another proto file. See also the `mgr_test.go` file. Or see my comment on the `mock.proto` file.

testing_gorums.go
```diff
@@ -37,3 +38,23 @@ func TestSetup(t testing.TB, numServers int, srvFn func(i int) ServerIface) ([]s
 	}
 	return addrs, stopFn
 }
+
+func TestServerSetup(t testing.TB, addr string, srv ServerIface) (func(), func()) {
```
meling, line 42: Maybe we should not export this func, since I think it is mainly aimed at testing gorums internal logic.

tests/mock/mock.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
```
meling, line 1: Consider reusing `tests/dummy/dummy.proto` instead; feel free to add message types to that file. Alternatively, you may rename it to `mock` if that is more appropriate. You should add/update the `tests/Makefile` to compile the proto file if `mock` is the name we go with.

meling (changes requested):

channel.go
```diff
@@ -200,23 +261,59 @@ func (c *channel) recvMsgs() {
 	}
 }
 
-func (c *channel) reconnect() {
-	c.streamMut.Lock()
-	defer c.streamMut.Unlock()
+func (c *channel) tryReconnect() {
```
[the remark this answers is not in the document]
meling: I wonder if some of the method names should be named something with `stream`. I wonder if it would be a clearer logical separation if we kept the logic of dialing to get a connection object as methods on the `RawNode` struct, and the stream-related "objects" in the `channel` struct.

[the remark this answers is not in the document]
aleksander-vedvik: channel.tryConnect() was initially created as a safety measure, but I agree that it is now more confusing than helping. The logic in channel.connect() could thus be moved to newChannel() instead. Then we could rename channel.tryConnect() to channel.connect(). This is not problematic as both newChannel() and channel.connect() is only used node.connect().

I don't think it is trivial to move the atomic flag to node.go because it needs to be checked every time a new message is sent. Either it would require a large refactor or that the channel is calling several methods on the node. In my opinion it would make more sense to move the dial method to the channel.

I have tried to rename some of the methods in the last commit to make it more clear/less confusing.

PR #198, opened 2025-01-11
feat(Logging): implemented a slog wrapper

aleksander-vedvik:
The broadcast code heavily relies on the logging wrapper. Hence, it is included first.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -1,0 +1,110 @@
+package logging
+
+import (
+	"log/slog"
+	"time"
+)
+
+// LogEntry is used in slog with correct types and json mapping
+type LogEntry struct {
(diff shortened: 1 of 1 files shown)


meling (approved):
Thanks for the PRs :-)

`BroadcastID` and `MachineID` are the only "labels" with capitalization (first letter). Does this have a particular significance, or should it be lower-cased to match the other labels?

aleksander-vedvik:
> Thanks for the PRs :-)
> 
> 
> 
> `BroadcastID` and `MachineID` are the only "labels" with capitalization (first letter). Does this have a particular significance, or should it be lower-cased to match the other labels?
> 
> 

Nope, they can be lower-cased 👌

PR #199, opened 2025-01-11
Features/broadcast/datatypes 2

aleksander-vedvik:
This pull request build on top of: https://github.com/relab/gorums/pull/198

This is a preparation pr that includes the most important types in the broadcast implementation.

### DIFF
--- a/broadcast/dtos/dtos.go
+++ b/broadcast/dtos/dtos.go
@@ -1,0 +1,75 @@
+package dtos
+
+import (
+	"context"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"time"
+)
+
+type Msg interface {
+	GetBroadcastID() uint64
+	GetMethod() string
+	String() string
+}
+
+type BroadcastMsg struct {
+	Ctx        context.Context
+	Options    BroadcastOptions
+	OriginAddr string
+	Info       Info
+}
+
+func (msg *BroadcastMsg) GetBroadcastID() uint64 {
+	return msg.Info.BroadcastID
+}
+
+func (msg *BroadcastMsg) GetMethod() string {
+	return msg.Info.Method
+}
+
+func (msg *BroadcastMsg) String() string {
+	return "broadcast"
+}
+
+type ReplyMsg struct {
+	Info       Info
+	ClientAddr string
+	Err        error
+}
+
+func (r *ReplyMsg) GetBroadcastID() uint64 {
+	return r.Info.BroadcastID
+}
+
+func (r *ReplyMsg) GetMethod() string {
+	return "reply"
+}
+
+func (r *ReplyMsg) String() string {
+	return "reply"
+}
+
+type Info struct {
+	Message         protoreflect.ProtoMessage
+	BroadcastID     uint64
+	Method          string
+	Addr            string
+	OriginMethod    string
+	OriginDigest    []byte
+	OriginSignature []byte
+	OriginPubKey    string
+}
+
+type Client struct {
+	Addr    string
+	SendMsg func(timeout time.Duration, dto *ReplyMsg) error
+	Close   func() error
+}
+
+type BroadcastOptions struct {
+	ServerAddresses  []string
+	AllowDuplication bool
+	SkipSelf         bool
+	ProgressTo       string
+	RelatedToReq     uint64
+}
--- a/broadcast/errors/errors.go
+++ b/broadcast/errors/errors.go
@@ -1,0 +1,57 @@
+package broadcastErrors
+
+type BroadcastIDErr struct{}
+
+func (err BroadcastIDErr) Error() string {
+	return "wrong broadcastID"
+}
+
+type MissingClientReqErr struct{}
+
+func (err MissingClientReqErr) Error() string {
+	return "has not received client req yet"
+}
+
+type AlreadyProcessedErr struct{}
+
+func (err AlreadyProcessedErr) Error() string {
+	return "already processed request"
+}
+
+type ReqFinishedErr struct{}
+
+func (err ReqFinishedErr) Error() string {
+	return "request has terminated"
+}
+
+type ClientReqAlreadyReceivedErr struct{}
+
+func (err ClientReqAlreadyReceivedErr) Error() string {
+	return "the client req has already been received. The forward req is thus dropped."
+}
+
+type MissingReqErr struct{}
+
+func (err MissingReqErr) Error() string {
+	return "a request has not been created yet."
+}
+
+type OutOfOrderErr struct{}
+
+func (err OutOfOrderErr) Error() string {
+	return "the message is out of order"
+}
+
+type ShardDownErr struct{}
+
+func (err ShardDownErr) Error() string {
+	return "the shard is down"
+}
+
+type InvalidAddrErr struct {
+	Addr string
+}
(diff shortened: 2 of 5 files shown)


broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+type ClientReqAlreadyReceivedErr struct{}
+
+func (err ClientReqAlreadyReceivedErr) Error() string {
+	return "the client req has already been received. The forward req is thus dropped."
+}
+
+type MissingReqErr struct{}
+
+func (err MissingReqErr) Error() string {
+	return "a request has not been created yet."
```
meling, line 36: Not sure I understand when this could happen, but here is a shorter error message:
```go
// Add comment here with more context
return "request not created yet"
```
Don't use `.` in error messages, since they often get chained:

```text
failed to update user: malformed request
```

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type OutOfOrderErr struct{}
+
+func (err OutOfOrderErr) Error() string {
+	return "the message is out of order"
```
meling, line 42:
```go
return "request received out of order"
```
Is this referring to a `client` request?

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type ShardDownErr struct{}
+
+func (err ShardDownErr) Error() string {
+	return "the shard is down"
```
meling, line 48:
```go
// more context?
return "shard is down"
```

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type InvalidAddrErr struct {
+	Addr string
+}
+
+func (err InvalidAddrErr) Error() string {
+	return "provided Addr is invalid. got: " + err.Addr
```
meling, line 56:
```go
return fmt.Sprintf("invalid address: %s", err.Addr)
```

broadcast/dtos/dtos.go
```diff
@@ -0,0 +1,75 @@
...
+	"time"
+)
+
+type Msg interface {
+	GetBroadcastID() uint64
+	GetMethod() string
+	String() string
+}
+
+type BroadcastMsg struct {
```
meling, line 15: It would be quite useful with doc comments for structs, methods, fields etc. Especially, those fields that aren't common knowledge, such as `Info`, `Options`, `OriginAddr`.

broadcast/dtos/dtos.go
```diff
@@ -0,0 +1,75 @@
...
+	Ctx        context.Context
```
meling, line 16: Difficult to assess, since I don't know the use of the `Ctx` field here, but it is a antipattern to include `context.Context` in structs. See the [Context and Structs blog post](https://go.dev/blog/context-and-structs). It should be passed as the first argument to all functions/methods that need a context. But see the blog for concrete advice.

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
+package broadcastErrors
+
+type BroadcastIDErr struct{}
+
+func (err BroadcastIDErr) Error() string {
+	return "wrong broadcastID"
+}
+
+type MissingClientReqErr struct{}
```
meling, line 9: Would be nice to have doc comments on these errors too.

broadcast/dtos/dtos.go
```diff
@@ -0,0 +1,75 @@
...
+	OriginPubKey    string
+}
+
+type Client struct {
+	Addr    string
+	SendMsg func(timeout time.Duration, dto *ReplyMsg) error
+	Close   func() error
+}
+
+type BroadcastOptions struct {
```
meling, line 69: This question applies to the all types and fields in the PR: Is it intentional that all types and fields are exported? Are they needed outside the `broadcast`/`dtos` package?

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
+package broadcastErrors
```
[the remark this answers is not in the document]
meling: Note also: Since the package is declared as `broadcastErrors` but the file is saved in the `errors` folder, this will not compile.

meling (changes requested):
So I did a rather thorough review here...

If you prefer that I push suggested changes directly to the PR, rather than asking you to fix my nitpicking comments. That way, I could instead ask only the questions that need more coordination and clarification. Let me know what's the best way forward.

aleksander-vedvik:
> So I did a rather thorough review here...
> 
> 
> 
> If you prefer that I push suggested changes directly to the PR, rather than asking you to fix my nitpicking comments. That way, I could instead ask only the questions that need more coordination and clarification. Let me know what's the best way forward.

Yes, please do. I'll instead update the subsequent PR's with the changes. 👍

broadcast/dtos/dtos.go
```diff
@@ -0,0 +1,75 @@
+package dtos
```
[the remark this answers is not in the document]
aleksander-vedvik: It's supposed to contain all dto's that are referenced outside of the broadcast package context (including subfolders). The main idea is to prevent cyclic imports and provide a common namespace for dto's pertaining to broadcast outside of the broadcast context. I.e. instead of referencing processor.RequestDto and router.Msg in gorums/server.go, I thought it would be easier to use dtos.RequestDto and dtos.Msg. Hoping this would reduce the cognitive load of developers working with higher level functionality such as nodes/channels, as both processor and router are implementation details specific to broadcasting.

Naming things are not my strong suit, so feel free to rename/change. If you think it is better, the dtos could also be moved to their respective packages (RequestDto -> processor, the rest -> router).

broadcast/errors/errors.go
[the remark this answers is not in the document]
aleksander-vedvik: My bad. The package should've been renamed to errors. I agree that they should not be in their own package. I'm just not very happy with how the logging wrapper currently operates, as it pollutes the code a bit (see the log statements in processor.go). I was hoping to improve this my including more context/fields in each error and log the errors directly. Hopefully making the logging a bit less verbose and at the same time making the errors a bit more explicit.

Since this is not implemented yet, it is probably better to move the errors to the packages that references them.

broadcast/dtos/dtos.go
aleksander-vedvik: Yes, they are. I provided a longer explanation to one of the comments above. 👍

broadcast/errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+type ReqFinishedErr struct{}
+
+func (err ReqFinishedErr) Error() string {
+	return "request has terminated"
+}
+
+type ClientReqAlreadyReceivedErr struct{}
+
+func (err ClientReqAlreadyReceivedErr) Error() string {
+	return "the client req has already been received. The forward req is thus dropped."
```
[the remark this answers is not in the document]
aleksander-vedvik: I like number 2 the best. 👍

aleksander-vedvik:
Most of the suggestions have been fixed. However, some requires major refactor (such as moving the errors). Is it possible to leave them as they are, and instead fix them in later pull requests?

PR #206, opened 2025-01-29
Added editions 2023 support with hybrid protobuf api

abjeni:
- **added editions support and updated storage example to edition2023**
Editions 2024 is still experimental, protoc has support but it is unclear if protoc-gen-go has support for it.

- **moving hybrid protobuf api**
The hybrid API allows for use of both the open and opaque protobuf api methods, and struct fields.

- ** Changes **
This may not have any effect for the users of gorums, i consider testing the opaque api on relab/raft. 
The gorums generated files have not changed.
The example worked both before and after being updated to the opaque API


abjeni:
benchmark/qspec.go line 61 is wrong.

Makefile
```diff
@@ -29,13 +30,17 @@ $(static_file): $(static_files)
 	@protoc-gen-gorums --bundle=$(static_file)
 
 %.pb.go : %.proto
-	@protoc -I=$(proto_path) --go_out=paths=source_relative:. $^
+	@protoc -I=$(proto_path) \
+		--go_opt=default_api_level=API_HYBRID \
```
meling, line 34: Nit: looks like an extra tab character on this line.

benchmark/qspec.go
```diff
@@ -42,26 +42,26 @@ func (qspec *QSpec) StopServerBenchmarkQF(_ *StopRequest, replies map[uint32]*Re
...
+	resp.SetLatencyAvg(resp.GetLatencyAvg() / float64(resp.GetTotalOps()))
 	for _, reply := range replies {
-		resp.LatencyVar += float64(reply.TotalOps-1) * reply.LatencyVar
+		resp.SetLatencyVar(resp.GetLatencyVar() + float64(reply.GetTotalOps()-1)*reply.GetLatencyVar())
 	}
-	resp.LatencyVar /= float64(resp.TotalOps) - float64(len(replies))
-	resp.TotalOps /= uint64(len(replies))
-	resp.TotalTime /= int64(len(replies))
-	resp.Throughput /= float64(len(replies))
+	resp.SetLatencyVar(resp.GetLatencyVar()/float64(resp.GetTotalOps()) - float64(len(replies)))
```
meling, line 61: Referring to your comment related to this line. I think it should be like this:
```go
resp.SetLatencyVar(resp.GetLatencyVar() / (float64(resp.GetTotalOps()) - float64(len(replies))))
```

meling (approved):
Very nicely done! Please fix the minor comments, and we can merge this.

One question I have is whether or not to update some or all the `.proto` files in `tests` and `internal/testprotos`, etc., to use editions. We can do it in a separate PR if this makes sense to do.

meling:
We would also need to merge master into this and resolve the merge conflicts, and we should fix the DeepSource issues as well.

meling (approved):
Looks good to me!

PR #229, opened 2025-05-06
register benchmark server for standalone servers

### LINKED ISSUE #228: bug: benchmark runServer does not register benchmark server
This is how the benchmark server is created:
```go
func runServer(server string, serverBuffer uint) {
	signals := make(chan os.Signal, 1)
	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)

	lis, err := net.Listen("tcp", server)
	checkf("Failed to listen on '%s': %v", server, err)

	srv := gorums.NewServer(gorums.WithReceiveBufferSize(serverBuffer))
	go func() { checkf("serve failed: %v", srv.Serve(lis)) }()

	fmt.Printf("Running benchmark server on '%s'\n", server)

	<-signals
	srv.Stop()
}
```

it does not use the RegisterBenchmarkServer that lets the server respond to rpc calls from clients, using benchmark.NewBenchServer in place of gorums.NewServer fixes it.

```go
func runServer(server string, serverBuffer uint) {
	signals := make(chan os.Signal, 1)
	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)

	lis, err := net.Listen("tcp", server)
	checkf("Failed to listen on '%s': %v", server, err)

	srv := benchmark.NewBenchServer(gorums.WithReceiveBufferSize(serverBuffer))
	go func() { checkf("serve failed: %v", srv.Serve(lis)) }()

	fmt.Printf("Running benchmark server on '%s'\n", server)

	<-signals
	srv.Stop()
}
```

abjeni:
Fixes #228

fixes the benchmark to register the server with the benchmark handlers.

### DIFF
--- a/cmd/benchmark/main.go
+++ b/cmd/benchmark/main.go
@@ -72,7 +72,7 @@ func runServer(server string, serverBuffer uint) {
 	lis, err := net.Listen("tcp", server)
 	checkf("Failed to listen on '%s': %v", server, err)
 
-	srv := gorums.NewServer(gorums.WithReceiveBufferSize(serverBuffer))
+	srv := benchmark.NewBenchServer(gorums.WithReceiveBufferSize(serverBuffer))
 	go func() { checkf("serve failed: %v", srv.Serve(lis)) }()
 
 	fmt.Printf("Running benchmark server on '%s'\n", server)
--- a/cmd/benchmark/main.go
+++ b/cmd/benchmark/main.go
@@ -217,7 +217,7 @@ func checkf(format string, args ...any) {
 	for _, arg := range args {
 		if err, _ := arg.(error); err != nil {
 			fmt.Fprintf(os.Stderr, format, args...)
-			os.Exit(1)
+			os.Exit(1) // skipcq: RVV-A0003
 		}
 	}
 }


abjeni:
consider only pulling the first commit.

benchmark/benchmark.go
```diff
@@ -13,14 +13,18 @@ import (
...
-	Remote     bool          // Whether the servers are remote (true) or local (false)
+	Concurrent     int           // Number of concurrent calls
+	Duration       time.Duration // Duration of benchmark
+	MaxAsync       int           // Max async calls at once
+	NumNodes       int           // Number of nodes to include in configuration
+	Payload        int           // Size of message payload
+	QuorumSize     int           // Number of messages to wait for
+	Warmup         time.Duration // Warmup time
+	Remote         bool          // Whether the servers are remote (true) or local (false)
+	ServerStats    bool
```
meling, line 24: Should add comments for these new fields.

benchmark/benchmark.go
```diff
@@ -13,14 +13,18 @@ import (
...
+	BenchReg       *regexp.Regexp
```
meling, line 25: Rename to `BenchRegExp` or just `RegExp`

cmd/benchmark/main.go
```diff
@@ -204,20 +216,27 @@ func main() {
 		CfgSize: options.NumNodes,
 	}
 
-	cfg, err := mgr.NewConfiguration(qspec, gorums.WithNodeList(remotes[:options.NumNodes]))
-	checkf("Failed to create configuration: %v", err)
+	cfg, err := mgr.NewConfiguration(qspec, gorums.WithNodeList(options.Remotes[:options.NumNodes]))
+	err = checkf("Failed to create configuration: %v", err)
```
meling, line 220: We should just drop the `checkf` thing and do:
```go
cfg, err := ...
if err != nil {
   return fmt.Errorf("Failed to create configuration: %w", err)
}
```
Or, alternatively just return the original error directly:
```go
if err != nil {
   return err
}
```
And similar for all the uses of `checkf` below in the `runClient` function. That is, we should only use `checkf` when we don't want to return an error but just exit if `err != nil`.

cmd/benchmark/main.go
```diff
@@ -204,20 +216,27 @@ func main() {
...
+		return err
+	}
 
-	printResults(results, options, *serverStats)
+	printResults(results, options)
+	return nil
 }
 
-func checkf(format string, args ...any) {
+func checkf(format string, args ...any) error {
```
meling, line 235: Should not return an error from this function since it is intended to exit on errors. If deep source is making a fuzz, let's just ignore those complaints. It you prefer, it is possible to add a `// skipcq: <Deep source code>` to force deep source to ignore it.

meling: I think the code might be:
```go
os.Exit(0) // skipcq: RVV-A0003
```

meling (changes requested):
Nice. A few smaller tweaks that should be fixed. I think we should just ignore complaints from deep source on these exit from non-main methods; see comments for specific advice.

meling (approved):

PR #234, opened 2025-06-25
Add support for server-side interceptors

### LINKED ISSUE #223: feat: add support for unary interceptors
We should add support for unary interceptors. The primary motivation for this is to support running multipaxos before executing a server handler; that is, only when a "command" has been decided by multipaxos should it execute the server handler. We should keep it in line with the standard gRPC interceptor style, perhaps even supporting chaining interceptors.

We can pass the interceptor as an option to `gorums.NewServer`:
```go
gorums.NewServer(gorums.UnaryInterceptor(p.unaryInterceptor)),
```

We could use the same or a similar function signature for the interceptor:
```go
func (k *xx) unaryInterceptor(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
	m, err := handler(ctx, req)
	if err != nil {
		return nil, err
	}
	return m, err
}
```

JosteinLindhom:
Closes #223

### DIFF
--- a/examples/storage/server.go
+++ b/examples/storage/server.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/relab/gorums"
+	"github.com/relab/gorums/examples/interceptors"
 	pb "github.com/relab/gorums/examples/storage/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
@@ -26,7 +27,11 @@ func startServer(address string) (*gorums.Server, string) {
 	storage.logger = log.New(os.Stderr, fmt.Sprintf("%s: ", lis.Addr()), log.Ltime|log.Lmicroseconds|log.Lmsgprefix)
 
 	// create Gorums server
-	srv := gorums.NewServer()
+	srv := gorums.NewServer(gorums.WithInterceptors(
+		interceptors.LoggingSimpleInterceptor,
+		interceptors.NoFooAllowedInterceptor[*pb.WriteRequest],
+		interceptors.MetadataInterceptor,
+	))
 	// register server implementation with Gorums server
 	pb.RegisterStorageServer(srv, storage)
 	// handle requests on listener


server.go
```diff
@@ -12,38 +12,44 @@ import (
...
+	}
+)
+
+func defaultInterceptor(ctx ServerCtx, msg *Message, next Handler) (*Message, error) {
+	// the default interceptor must be the outermost interceptor in the chain
+	// it releases the handler lock after the next handler returns
+	// allowing the next request to be processed
+	defer ctx.Release()
+	message, err := next(ctx, msg)
+	if message == nil && err == nil {
```
meling: Should we add a comment to explain what scenario we may experience `message==nil` and why we don't want to invoke `SendMessage` for this case?

server.go
```diff
@@ -136,10 +143,46 @@ func WithConnectCallback(callback func(context.Context)) ServerOption {
 	}
 }
 
+func WithInterceptors(i ...Interceptor) ServerOption {
```
meling:
```go
// WithInterceptors registers server-side interceptors to run for every incoming request.
// Interceptors are executed for each registered handler. Interceptors are executed in
// the order they are provided.
```

server.go
```diff
@@ -136,10 +143,46 @@ func WithConnectCallback(callback func(context.Context)) ServerOption {
...
+		curr := interceptors[i]
+		next := handler
+		handler = func(ctx ServerCtx, msg *Message) (*Message, error) {
+			return curr(ctx, msg, next)
+		}
+	}
+	return handler
+}
+
+// final = someHandler
```
meling: Remove comments.

server.go
```diff
@@ -12,38 +12,44 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
```
meling: Could you add in `server_test.go`:
- a test with a single interceptor that tests expected behavior
- a test with two interceptors that tests ordered changes (i1, i2, final, i2, i1): the first interceptor could just add the string `i1in-`, the second could add `i2in-`, the server method handler could add `final-`, the second interceptor could add `i2out-`, and finally, the first interceptor could add `i1out`. Then the check would be that the client got the response: `i1in-i2in-final-i2out-i1out`

meling (changes requested):

meling (changes requested):
@JosteinLindhom I've fixed most of my own comments, but there are a couple more that could benefit from your input. Moreover, please review my changes as well.

PR #244, opened 2025-11-26
Feat: Support client-side interceptors

### LINKED ISSUE #37: Redesign Gorums to use Client Interceptors
This would be a fairly large redesign, but if my initial thoughts are correct, we could possibly get rid of the whole code generator. I may be wrong of course.

Here is the godoc: 
https://godoc.org/google.golang.org/grpc#UnaryClientInterceptor

Here is a simple example of using a Client Interceptor to measure time:
https://about.sourcegraph.com/go/grpc-in-production/#client-interceptor
Same for a Server Interceptor:
https://about.sourcegraph.com/go/grpc-in-production/#server-interceptor
Here is a Client Interceptor for retry logic:
https://about.sourcegraph.com/go/grpc-in-production/#networks-fail

Here is a youtube video with more details (not only on inceptors):
https://www.youtube.com/watch?v=7FZ6ZyzGex0

Another source of reference examples can be found here:
https://github.com/grpc-ecosystem/go-grpc-middleware

meling:
This PR implements client-side interceptor logic to support an interceptor-based quorum call architecture. In a future update, I will update the `protoc-gen-gorums` generator to use the new interceptor architecture. This will significantly simplify our generator code, and allow us to remove several gorums options that will no longer be needed.

Implement a flexible, composable interceptor architecture for quorum calls that provides better type safety, modularity, and extensibility compared to the legacy `QuorumCall` approach.

## Core Architecture:

- `QuorumInterceptor`: Generic type for interceptor functions that wrap `QuorumFunc`
- `QuorumFunc`: Function signature for processing quorum calls and returning results
- `ClientCtx`: Context object providing access to request, config, and response iterator
- `Chain`: Utility function to compose interceptors around a base handler

## Key Features:

1. Full type safety with generics (Req, Resp, Out type parameters)
2. Support for custom return types via Out parameter
3. Lazy message sending - transforms applied before dispatch
4. Iterator-based response handling with early termination support
5. Composable middleware pattern for building complex quorum logic

## Base Quorum Functions (Aggregators):

- `MajorityQuorum`: Returns first response after ⌈(n+1)/2⌉ successful replies
- `FirstResponse`: Returns first successful response (read-any pattern)
- `AllResponses`: Waits for all nodes to respond (write-all pattern)
- `ThresholdQuorum`: Generic threshold-based quorum with configurable count
- `CollectAllResponses`: Returns a map of all successful responses by node ID

## Interceptors (Middleware):

- `PerNodeTransform`: Applies per-node request transformations with skip support
- `QuorumSpecInterceptor`: Adapter for legacy QuorumSpec-style functions (will be removed)

## Iterator Helpers:

- `IgnoreErrors`: Filters the iterator to yield only successful responses
- `CollectN`: Collects up to n responses into a map
- `CollectAll`: Collects all responses into a map

## Implementation Details:

- Lazy sending via sync.OnceFunc ensures transforms are registered before dispatch
- RegisterTransformFunc allows chaining multiple request transformations
- applyTransforms applies registered transforms in order, skips invalid results
- Responses() iterator yields node responses as they arrive
- Type-safe conversion from Result[proto.Message] to Result[Resp]

## Backward Compatibility:

- Legacy QuorumCall remains unchanged
- QuorumSpecInterceptor bridges old and new approaches
- No breaking changes to existing code

The new architecture enables gradual migration from the legacy approach; however, the plan is to replace it wholesale. Hence, there is no plan to support legacy quorum call types.

## Testing:

- 17 comprehensive test functions covering unit and integration scenarios
- Tests for iterator utilities, interceptor chaining, and custom aggregation
- Tests for per-node transformation with node skipping
- Tests for all base quorum functions with various error conditions
- Integration tests with real gRPC servers
- Helper functions for consistent test patterns (testContext, checkError, etc.)

## Updated API

After the initial design, I have updated the API to use composable iterator methods instead of the function-based variant promoted by the standard library. This was inspired by Asbjørn Salhus's implementation in #230, which was not designed for interceptors. The new design is now more composable.

Fixes #37

### DIFF
--- a/async.go
+++ b/async.go
@@ -37,7 +37,7 @@ func (f *Async) Done() bool {
 type asyncCallState struct {
 	md              *ordering.Metadata
 	data            QuorumCallData
-	replyChan       <-chan Result[proto.Message]
+	replyChan       <-chan NodeResponse[proto.Message]
 	expectedReplies int
 }
 
@@ -47,7 +47,7 @@ type asyncCallState struct {
 func (c RawConfiguration) AsyncCall(ctx context.Context, d QuorumCallData) *Async {
 	expectedReplies := len(c)
 	md := ordering.NewGorumsMetadata(ctx, c.getMsgID(), d.Method)
-	replyChan := make(chan Result[proto.Message], expectedReplies)
+	replyChan := make(chan NodeResponse[proto.Message], expectedReplies)
 
 	for _, n := range c {
 		msg := d.Message
--- a/channel.go
+++ b/channel.go
@@ -14,8 +14,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// Result wraps a response value from node ID, and an error if any.
-type Result[T any] struct {
+// NodeResponse wraps a response value from node ID, and an error if any.
+type NodeResponse[T any] struct {
 	NodeID uint32
 	Value  T
 	Err    error
@@ -28,7 +28,7 @@ type request struct {
 	msg          *Message
 	opts         callOptions
 	streaming    bool
-	responseChan chan<- Result[proto.Message]
+	responseChan chan<- NodeResponse[proto.Message]
 }
 
 type channel struct {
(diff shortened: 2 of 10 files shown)


meling:
Superceded by #245

